PR #20, opened 2022-02-27
Profile page

anubra266:
Fixes EveripediaNetwork/issues#84


kesar:
Captura de pantalla 2022-03-03 a las 12 39 24

hero image should not have extra space w the header

Captura de pantalla 2022-03-03 a las 12 39 24

lets make sure the have tooltips like the opensea one and lets add the settings one that should be visible if you access to your own profile. (we can leave the detect user profile for a later issue, but lets at least have the button there)

Captura de pantalla 2022-03-03 a las 12 40 39

anubra266:
> Captura de pantalla 2022-03-03 a las 12 36 55
> 
> footer is broken in my laptop

Oh @kesar That's the fixed sidebar staying over the footer

kesar:
> @kesar I think that's why opensea doesn't have a sidebar there. I can't think of any UX where we have that sidebar fixed and still have the footer. And I don't think we need footers in any page that's authenticated. I don't think it's done anywhere too.

Interesting, I didnt know about that pattern, its probably right. Let's make the footer optional so we can turn it on / off depending of the page. and lets make sure its off in profile 👍🏻 

I will open new issue to change it in other pages once this gets merged

anubra266:
@kesar can we merge this

kesar (approved):

PR #47, opened 2022-03-04
Fix category page

Royal-lobster:
Fixes "https://github.com/EveripediaNetwork/issues/issues/101"

### DIFF
--- a/src/pages/categories/[category].tsx
+++ b/src/pages/categories/[category].tsx
@@ -49,7 +49,7 @@ const Category: NextPage = () => {
           my={12}
           gap={8}
         >
-          {Array.from(new Array(6), (_x, i) => i).map(i => (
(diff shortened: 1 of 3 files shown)


src/pages/categories/[category].tsx
```diff
@@ -48,7 +49,7 @@ const Category: NextPage = () => {
           my={12}
           gap={8}
         >
-          {[...Array(6)].map(i => (
+          {Array.from(new Array(6), (_x, i) => i).map(i => (
```
Comment (anubra266) [inline src/pages/categories/[category].tsx:52]: You can do:
```suggestion
          {Array.from({ length: 6 }).map((_, i) => (
```
More readable and involves less declarations

Comment (Royal-lobster) [reply src/pages/categories/[category].tsx]: Awesome !

kesar (approved):

PR #57, opened 2022-03-08
Categories Integration with API

Royal-lobster:
[Draft] Closes https://github.com/EveripediaNetwork/issues/issues/128


kesar:
take a look to: https://github.com/EveripediaNetwork/ep-ui/tree/main/src/services/wikis to save graphql queries into the store

kesar:
i think if we use an approach similar to: https://github.com/EveripediaNetwork/ep-ui/blob/main/src/pages/wiki/%5Bslug%5D.tsx#L31

we can get categories from SSR which would improve speed (hopefully) and SEO

Royal-lobster:
> i think if we use an approach similar to: https://github.com/EveripediaNetwork/ep-ui/blob/main/src/pages/wiki/%5Bslug%5D.tsx#L31
> 
> & https://github.com/EveripediaNetwork/ep-ui/blob/main/src/pages/wiki/%5Bslug%5D.tsx#L17
> 
> we can get categories from SSR which would improve speed (hopefully) and SEO

Ohh that's awesome ! Thanks for pointing !

kesar:
btw this needs to be done also for the categories menu in navbar and for the selection in the create wiki page

we should also remove the dummy old data once its done 👍🏻

src/components/Layout/Navbar/DesktopNav.tsx
```diff
@@ -3,10 +3,40 @@ import { HStack } from '@chakra-ui/react'
...
   const [visibleMenu, setVisibleMenu] = useState<number | null>(null)
+  const [navData, setNavData] = useState<NavItem[]>(NAV_ITEMS)
+  const { data: topCategoriesLinks } = useGetTopCategoriesLinksQuery()
+
+  // Update nav data when top categories links are loaded
+  useEffect(() => {
+    if (topCategoriesLinks) {
+      const topCategories = topCategoriesLinks.map(
+        ({ title, id, icon }, i) => ({
+          id: parseInt(`10${i}${2}`, 10),
```
kesar, line 20: why is this? :O

src/pages/categories/index.tsx
```diff
@@ -1,46 +1,66 @@
...
+        <SimpleGrid
+          columns={{ base: 1, sm: 2, lg: 3 }}
+          width="min(90%, 1200px)"
+          mx="auto"
+          my={12}
+          gap={8}
+        >
+          {data?.map((category, i) => (
+            <CategoryCard
+              key={category.id || sampleCategories[i].id}
```
kesar, line 44: no need to keep compatibility from sample

src/components/Layout/Navbar/MobileNav.tsx
```diff
@@ -39,6 +40,34 @@ const MobileNav = ({ toggleWalletDrawer, setHamburger }: MobileNavType) => {
...
+  // Update nav data when top categories links are loaded
+  useEffect(() => {
+    if (topCategoriesLinks) {
+      const topCategories = topCategoriesLinks.map(
+        ({ title, id, icon }, i) => ({
+          id: parseInt(`10${i}${2}`, 10),
+          label: title || id,
+          href: `/categories/${id}`,
+          hasImage: true,
+          icon,
```
kesar, line 55: I am expecting that Icon from the API will be a string like "BsBriefcaseFill" and then we will use that to put the Icon in the navbar. Not sure if that's how its working in this PR

kesar:
image


I added more dummy data into the db that way we can see the loading of images / icons / etc 👍🏻

src/components/Layout/Navbar/MobileNav.tsx
Royal-lobster: Yes i am wondering how the icons are handled so i left it blank 😅. The string would work well with dynamic import so i will implement it in next commit !

src/pages/categories/index.tsx
Royal-lobster: ohh yes. before i implemented the ui with slugs so i changed the implementation to use IDs. just wanted to test it out so i modified the sample data temporary.

src/components/Layout/Navbar/DesktopNav.tsx
Royal-lobster: I was playing with RTK query 😅 the query fetches the top 10 categories from the database and injections in nav items data on client side to show in UI. That should be done on serverside using SSR so i am searching for elegant way to implement it. that should be sorted in my next commit 😅

src/data/CategoriesData.ts
```diff
@@ -15,7 +15,7 @@ import {
 
 export const sampleCategories: CategoryDataType = [
```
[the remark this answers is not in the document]
Royal-lobster: Yes. we can remove it after fully integrating with API.

Royal-lobster:
> image
> 
> I added more dummy data into the db that way we can see the loading of images / icons / etc 👍🏻

That's awesome ! makes it easier to play and integrate ! Thanks 😊

src/data/CategoriesData.ts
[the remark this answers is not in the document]
kesar: I think we can remove it before merging right? (in this PR)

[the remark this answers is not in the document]
Royal-lobster: Yes ! i also changed usage of sample data inside create-wiki in highlight modal to API and deleted the sample data. Now its fully integrated

kesar (approved):

PR #61, opened 2022-03-12
Trail to fix API request flood

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/133

It seems like the accountData is getting triggered recursively inside useEffect in `src/components/Layout/WalletDrawer/Connectors.tsx` I updated the dependencies to be more specific that are used in useEffect body

## Before

https://user-images.githubusercontent.com/52039218/158016519-a0107614-26bc-460b-bccc-1d7bda260e13.mp4

## After 

https://user-images.githubusercontent.com/52039218/158016527-9cf02704-a029-4d43-ba94-2bbbbca172db.mp4

### DIFF
--- a/src/components/Layout/WalletDrawer/Connectors.tsx
+++ b/src/components/Layout/WalletDrawer/Connectors.tsx
@@ -45,7 +45,7 @@ const Connectors = () => {
     useState<boolean>(true)
 
   useEffect(() => {
-    if (address && !walletDetails && accountData) {
+    if (address && !walletDetails && accountData?.address) {
       const payload = {
         address,
         connector: undefined,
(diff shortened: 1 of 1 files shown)


kesar:
that's a great catch. Can you check the reason why once you are logged and refresh the page it does around 130 request? it sounds like too many requests. Some seems to be duplicated, some maybe can be cached somehow (perhaps its an issue in the lib we are using itself).

It also seems to be using many different endpoints from different APIs that might be they are duplicating work. It would be worth to analyze what's going on there 👍🏻 

image

image

Royal-lobster:
> that's a great catch. Can you check the reason why once you are logged and refresh the page it does around 130 request? it sounds like too many requests. Some seems to be duplicated, some maybe can be cached somehow (perhaps its an issue in the lib we are using itself).
> 
> It also seems to be using many different endpoints from different APIs that might be they are duplicating work. It would be worth to analyze what's going on there 👍🏻
> 
> image
> 
> image

Yes ! I will try to figure out why its happening

kesar (approved):

PR #62, opened 2022-03-13
Add iq to metamask

Emmanueldmlr:
fixes https://github.com/EveripediaNetwork/issues/issues/108


Emmanueldmlr:
@kesar kindly review.

src/components/Layout/Token/TokenDetailsMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {
+          method: 'metamask_watchAsset',
+          params: {
+            type: 'ERC20',
+            options: {
+              address: config.iqAddress,
+              symbol: tokenSymbol,
+              decimals: 18,
+              image:
+                'https://pbs.twimg.com/profile_images/1414736076158033921/nResATsF_400x400.png',
```
kesar, line 25: this is better if we self-hosted it in our local

src/components/Layout/Token/TokenDetailsMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+            },
+          },
+          id: Math.round(Math.random() * 100000),
```
kesar, line 28: is this id needed? and it has to be randomized?

src/components/Layout/Token/TokenDetailsMenu.tsx
Emmanueldmlr: Okay, I will switch it to local then.

src/components/Layout/Token/TokenDetailsMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  return (
+    <Menu placement="left-start">
+      <MenuButton>
+        <RiMore2Fill color="color" fontSize="20" fontWeight="bold" />
+      </MenuButton>
+      <MenuList onClick={() => handleAddTokenToMetamask(token)} boxShadow="xl">
+        <MenuItem>Add {token} token to Metamask</MenuItem>
```
kesar, line 43: I think we need a whitelist of tokens here. ETH doesnt work for example, so probably we should hide the 3 dots menu from the ones that can't be added 

I would just have this for IQ for now 👍🏻

kesar:
functionality tested on IQ works fine 👍🏻 just some minor suggestions

src/components/Layout/Token/TokenDetailsMenu.tsx
Emmanueldmlr: It's Optional. I think I should remove it then?

kesar: yeah, if it works without id, probably its not worth to add it (unless we know the reason behind 👍🏻 )

src/components/Layout/Token/TokenDetailsMenu.tsx
Emmanueldmlr: Okay. I will do that now. Thanks for the review.

src/components/Layout/Token/TokenDetailsMenu.tsx
Emmanueldmlr: Let me try it out

src/components/Layout/Token/TokenDetailsMenu.tsx
Emmanueldmlr: @kesar, hiding the 3dots will affect the UI. so, instead, I disabled it (3 dots) when the token is not part of the whitelist of tokens.

kesar: sounds good 👍🏻

PR #74, opened 2022-03-19
Category wikis

Emmanueldmlr:
fixes https://github.com/EveripediaNetwork/issues/issues/147


package.json
```diff
@@ -47,6 +47,7 @@
     "ipfs-http-client": "55.0.0",
     "ismobilejs": "^1.1.1",
     "match-sorter": "^6.3.1",
+    "moment": "^2.29.1",
```
kesar, line 50: https://blog.bitsrc.io/date-fns-vs-momentjs-9833f7463751

pls use date fns if possible

Emmanueldmlr: Okay then, I will update the codebase and the package.json file. Thanks for the resource, what an eye-opener.

src/components/Categories/SubCategoryCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Box
+        w={390}
+        minH={390}
+        boxShadow="2xl"
+        rounded="md"
+        p={6}
+        overflow="hidden"
+      >
+        <Box h={200} mb={3} pos="relative">
+          <Image src="/images/sub-category-image.png" layout="fill" />
```
kesar, line 22: image must come from wiki and default to a default one if image returns 404

Emmanueldmlr: image
Images array doesn't have an Image property on it

kesar: https://gateway.pinata.cloud/ipfs/ + id

first part of the url i think its an env / config var

Emmanueldmlr: Wao...I didn't know about this. Thanks. I will implement it now.

kesar (approved):

PR #77, opened 2022-03-21
Wikis API Migrate and Wiki Type Change

Royal-lobster:
# Changes
1. Changes wiki RTK query to use Graphql API instead of theGraph 
2. ⚠️ Changes the type of ```Wiki``` to match the data from backend
3. ⚠️ Changes the structure of data stored in ```wiki.slice.ts``` to match new wiki type
4. ⚠️ Made Changes at all places to accommodate the new wiki type that uses wiki slice or wiki RTK query 

## Difference between old and new wiki type
basically spreads the old contents object in to wiki type

screely-1647865123673

This removes lot of destructing inside destructing for content.

### DIFF
--- a/src/components/Layout/Editor/Highlights/Highlights.tsx
+++ b/src/components/Layout/Editor/Highlights/Highlights.tsx
@@ -28,7 +28,7 @@ const Highlights = () => {
   const handleDispatch = (object: Partial<Content>) =>
     dispatch({
       type: 'wiki/setCurrentWiki',
-      payload: { content: { ...object } },
+      payload: object,
     })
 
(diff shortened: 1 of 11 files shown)


Royal-lobster:
I tried to find and patch the new changes where its needed but i am not 100% sure if the patches are correct.
@kesar @JustinCast @Emmanueldmlr Kindly review the changes

kesar:
for me it looks good as well. as long as it does not change the payload we send in create wiki i think everything else is fine 👍🏻

kesar:
merging this so we can work in edit wiki without conflicts 👍🏻

PR #84, opened 2022-03-23
hook up profile to API

anubra266:
Resolves EveripediaNetwork/issues#144

CleanShot 2022-03-23 at 20 02 10@2x
CleanShot 2022-03-23 at 20 02 17@2x


kesar:
I'm going to merge everything for today's demo and then we can work in specifics 👍🏻

PR #94, opened 2022-03-25
Feat/edit page

JustinCast:
# Edit page
_Uses the same layout as create a new wiki, but it instead verifies if there is a slug and then request the wiki data._

## How should this be tested?
1. `yarn start`
2. Go to the attached link/create-wiki?slug=slug-name-goes-here
3. Wait for the data to be fetched

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/163


src/components/Elements/Dropzone/Dropzone.tsx
```diff
@@ -45,6 +47,11 @@ const Dropzone = ({ dropZoneActions }: DropzoneType) => {
     accept: 'image/jpeg, image/png, image/jpg',
   })
 
+  useEffect(() => {
+    if (initialImage)
+      setPaths([`https://ipfs.everipedia.org/ipfs/${initialImage}`])
```
kesar, line 52: ipfs url has a config value I think (if not pls add it)

src/pages/_app.tsx
```diff
@@ -38,9 +39,11 @@ const App = (props: EpAppProps) => {
         <ChakraProvider resetCSS theme={chakraTheme}>
           <Fonts />
           <Provider autoConnect connectors={connectors}>
-            <Layout noFooter={Component.noFooter}>
-              <Component {...pageProps} />
-            </Layout>
+            <ImageProvider>
```
kesar, line 42: shouldnt this provider be at wiki create page level? I dont see any other use case outside of that scope

kesar (changes requested):
two small changes, besides that it looks good

src/pages/_app.tsx
JustinCast: > shouldnt this provider be at wiki create page level? I dont see any other use case outside of that scope

I had to include the image provider in the _app.tsx file, since in the nested component wasn't working

PR #95, opened 2022-03-27
Fix local storage issue

Emmanueldmlr:
fixes https://github.com/EveripediaNetwork/issues/issues/184

@kesar , I'm only storing the user details in the local storage for persistence now. Since the main reason why we added local storage is just to maintain the session.

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -17,8 +17,8 @@ import { store } from '@/store/store'
 import { getCategoriesLinks } from '@/services/categories'
 import { getRunningOperationPromises } from '@/services/wikis'
 import Fonts from '@/theme/Fonts'
-import chakraTheme from '../theme'
 import { saveState } from '@/utils/browserStorage'
+import chakraTheme from '../theme'
 
 type EpAppProps = AppProps & {
   Component: AppProps['Component'] & { noFooter?: boolean }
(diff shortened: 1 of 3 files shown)


src/store/store.ts
```diff
@@ -25,7 +25,7 @@ export const store = configureStore({
     [navSearchApi.reducerPath]: navSearchApi.reducer,
   },
   middleware: gDM =>
-    gDM({ serializableCheck: true })
+    gDM({ serializableCheck: false })
```
kesar, line 28: why remove this? i think it give us warnings when we put in storage data that we shouldnt

src/pages/_app.tsx
```diff
@@ -27,7 +28,7 @@ const App = (props: EpAppProps) => {
 
   store.subscribe(
     debounce(() => {
-      // saveState(store.getState())
+      saveState(store.getState())
```
kesar, line 31: perhaps we can simply not use localstorage for redux and just store and retrieve account name in a more traditional way.

This is executing code each 800ms for a reason that perhaps we dont need.

kesar: something similar to: https://github.com/EveripediaNetwork/ep-ui/pull/93/files

src/store/store.ts
Emmanueldmlr: I changed the status because we are only storing the user object in the locastorage which is confirmed that it is serializable. Leaving it to be true is causing the error in the console because its seeing providerSlice as not serializabke in the store

src/pages/_app.tsx
kesar: we can skip the provider in the store and just store in redux stuff that its serializable

Emmanueldmlr: This will work then.

PR #134, opened 2022-04-02
Social Cards for Wiki and Abstraction of wiki image link and summary

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/195


src/pages/wiki/[slug].tsx
```diff
@@ -67,27 +70,44 @@ const Wiki = () => {
...
-            {wiki && <WikiInsights wiki={wiki} />}
+    <>
+      <NextSeo
+        title={wiki?.title}
+        openGraph={{
+          title: wiki?.title,
+          description: shortenText(wiki?.content || '', 180),
+          images: [
+            {
+              url: `${config.pinataBaseUrl}${
```
Comment (kesar) [inline src/pages/wiki/[slug].tsx:81]: a good refactoring here would be to create a function in utils that expect a partial wiki that can have images and return an url (and replace it everywhere)

that way if we ever need to switch to something like a proxy url that generates the image w some optimizations we only need to change it in one place 👍🏻

Comment (Royal-lobster) [reply src/pages/wiki/[slug].tsx:81]: Ohh that will be great ! i will do that.

Comment (kesar) [reply src/pages/wiki/[slug].tsx:81]: we could actually do the same w description, and use that for wiki cards descriptions / search bars and then, when we do the refactoring to replace that for a summary, we only need to change it once 👍🏻

kesar:
lets make sure we replace description & image across all the site 👍🏻

src/components/Categories/SubCategoryCard.tsx
```diff
@@ -30,7 +30,7 @@ const SubCategoryCard = ({ wiki }: { wiki: Wiki }) => {
               {title}
             </Text>
             <Text color="gray.600" fontSize="md">
-              {shortenText(content, 65)}
+              {getWikiSummary(wiki, 65)}
```
kesar, line 33: extra points: we can avoid magic numbers and all readability:

export enum WikiSummarySize {
  Small = 65,
  Medium = 70,
  Big = 160
};

const size: WikiSummarySize = WikiSummarySize .Small;

Royal-lobster: Ohh that's will be awesome ! i will implement that

src/utils/getWikiImageUrl.ts
```diff
@@ -1,6 +1,7 @@
 import config from '@/config'
+import { WikiTitle } from '@/services/nav-search'
 import { Wiki } from '@/types/Wiki'
 
-export const getWikiImageUrl = (wiki: Partial<Wiki>) => {
+export const getWikiImageUrl = (wiki: Partial<Wiki> | WikiTitle) => {
```
kesar, line 5: not needed for this PR but something would increase quality its to work out the types. I think that WikiTitle and this partial could be replaced w something that its partial and require specific properties:

https://stackoverflow.com/a/48244432

we can rework the types later on, but just to keep that in mind, that sometimes we create types that could be partials of the main Wiki with some expected properties 👍🏻

Royal-lobster: Wow i don't know about that. Thanks for the resource 😄

kesar (approved):

kesar (approved):

PR #152, opened 2022-04-07
Fixes editor auto focus scrolling page to bottom

Royal-lobster:
Re-implemented fetching initial wiki content to editor. Did this for using `autofocus={false}` directly on `ToastUIEditor` component.

### DIFF
--- a/src/components/Layout/Editor/Editor.tsx
+++ b/src/components/Layout/Editor/Editor.tsx
@@ -9,49 +9,24 @@ import wikiLink from '@/editor-plugins/wikiLink'
 
 type EditorType = {
   onChange: (value: string | undefined) => void
-  initialValue: string
   markdown: string
 }
 
-const Editor = ({ onChange, initialValue, markdown }: EditorType) => {
+const Editor = ({ onChange, markdown }: EditorType) => {
   const { colorMode } = useColorMode()
   const editorRef = useRef<ToastUIEditor>(null)
   const containerRef = useRef<HTMLDivElement>(null)
 
-  const callEditorMethod = useCallback(() => {
-    const currentMarkdown = editorRef.current
-      ?.getInstance()
-      .getMarkdown()
-      .toString() as string
-
-    if (currentMarkdown !== markdown)
-      editorRef.current?.getInstance()?.setMarkdown(markdown)
-  }, [editorRef.current?.getInstance()?.getMarkdown(), markdown])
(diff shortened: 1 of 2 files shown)


Royal-lobster:
@JustinCast @Emmanueldmlr Please kindly review. I am not 100% sure if this change don't break any other functionality as side-effect.

src/components/Layout/Editor/Editor.tsx
```diff
@@ -61,20 +36,20 @@ const Editor = ({ onChange, initialValue, markdown }: EditorType) => {
     if (markdown !== currentMd) onChange(currentMd)
   }, [editorRef.current?.getInstance().getMarkdown(), markdown, onChange])
 
-  useEffect(() => {
-    updateEditorHeaderBackground(colorMode)
-  }, [colorMode])
-
+  console.log(markdown)
```
JustinCast, line 39: we can remove this console.log

JustinCast (changes requested):
The editor text is not changing after change the page type

src/components/Layout/Editor/Editor.tsx
kesar, line 39:
```suggestion
```

when there is something simple like this I would recommend to click in suggest button and send the suggestion since it takes no time and PR could even get accepted without getting back to the IDE if the creator commit suggestion 👍🏻 cc: @JustinCast

kesar, line 39:
```suggestion
```

when there is something simple like this I would recommend to click in suggest button and send the suggestion since it takes no time and PR could even get accepted without getting back to the IDE if the creator commit suggestion 👍🏻 cc: @JustinCast

Royal-lobster:
> The editor text is not changing after change the page type

Ohh i was thinking the same but, The editor text is not changing in main branch too.

kesar:
> > The editor text is not changing after change the page type
> 
> Ohh i was thinking the same but, The editor text is not changing in main branch too by changing the page type.

if its a issue on main, please open an issue related to that on github and we could merge this I guess 👍🏻

Royal-lobster:
Created a new issue for that !

kesar (approved):

kesar:
cool, lets wait for @JustinCast input 👍🏻

JustinCast (approved):
LGTM

PR #154, opened 2022-04-07
FIx profile image

anubra266:
**Before**
CleanShot 2022-04-07 at 21 18 39@2x

**After**
CleanShot 2022-04-07 at 21 24 23@2x


kesar (approved):
sorry about that! I removed it before the meeting to get other part of the app working! Thanks for the fix

PR #157, opened 2022-04-08
Implementing twitter widget.

codiejay:
Users can now add, optionally, their twitter url. This will allow a timeline of their twt account to appears on the wiki page

## Linked issues

closes #208, closes #208, closes https://github.com/EveripediaNetwork/issues/issues/208


codiejay:
> if i go to edit a page to add twitter, it doesnt work

@kesar This is cause when those wikis where created they didn't have the `Twitter profile` key attached to them. But right now if you create a wiki on the branch it has the twitter profile option and you can also edit the twitter profile.

kesar:
> > if i go to edit a page to add twitter, it doesnt work
> 
> 
> 
> @kesar This is cause when those wikis where created they didn't have the `Twitter profile` key attached to them. But right now if you create a wiki on the branch it has the twitter profile option and you can also edit the twitter profile. 

old wikis should allow to add new metadata like twitter or any other future metadata

kesar:
also lets ask for twitter handle not twitter url, so just kesarito for example instead of https://twitter.com/kesarito.


image

also if i add a twitter handle, close dialog, and open it again, input should be populated w last value

kesar (changes requested):
https://github.com/EveripediaNetwork/ep-ui/pull/157#issuecomment-1093060404
https://github.com/EveripediaNetwork/ep-ui/pull/157#issuecomment-1094918369

kesar (approved):

PR #158, opened 2022-04-08
fix: pin the image using the backend service

JustinCast:
# Use the backend service to pin the wiki main image
_Use the backend service to pin the main wiki image instead of having a separate functionality for that in the nextjs api_

## How should this be tested?
- Using the attached link

## Linked issues
closes https://github.com/EveripediaNetwork/issues/issues/229


kesar (approved):

kesar (changes requested):
now that we are fully removing ipfs from nextjs i think there are some packages that can be removed from our packages.json

JustinCast:
> now that we are fully removing ipfs from nextjs i think there are some packages that can be removed from our packages.json

let me take a look at it

kesar (approved):

PR #160, opened 2022-04-09
fixes weird redirect from edit wiki

Emmanueldmlr:
# Weird redirection from edit wiki to create wiki after login

_ sometimes, when a user visits the wiki edit page and there is a need to re-authenticate the user, the user is redirected to the login page for authentication and after authentication, the user is redirected to the create wiki page instead of the wiki edit page. The redirection skips the query parameters. 

## How should this be tested?

1. Navigate to your profile page to have access to your already created wikis.
2. Click on one of your wikis to view it.
3. Click on the edit wiki button to edit your wiki
4. If you are redirected to the login page for authentication, after authentication, you must be redirected back to the wiki edit page with the wiki content in place.

fixes https://github.com/EveripediaNetwork/issues/issues/230


src/pages/login/index.tsx
```diff
@@ -8,6 +8,9 @@ import { useRouter } from 'next/router'
 const Login = () => {
   const { user } = useSelector((state: RootState) => state.user)
   const router = useRouter()
+  console.log(router.query.from)
+  console.log(router.query.from)
+  console.log(router.query.from)
```
kesar, line 13:
```suggestion
```

src/components/AuthenticatedRoute/index.tsx
```diff
@@ -11,11 +12,14 @@ export const authenticatedRoute = <P extends object>(
...
       useSelector((state: RootState) => state.user.user) || getState()
+
     useEffect(() => {
       if (!user) {
+        const previousPageQuery = router.query as GetUrlQueriesType
+        const queryAfterLogin = getUrlQueries(previousPageQuery) || ''
         router.push({
           pathname: '/login',
-          query: { from: router.pathname },
+          query: { from: `${router.pathname}${queryAfterLogin}` },
```
kesar, line 22: shouldnt this be fixed using asPath? https://nextjs.org/docs/api-reference/next/router

Emmanueldmlr: I still can't believe that Nextjs has a faster way of achieving that. I feel bad right now...😭😭😭😭

kesar (approved):

PR #171, opened 2022-04-15
Wiki preview on link hover

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/227

# Changes
1. Added wiki link preview in wiki page
2. Made type ```WikiTitle```  -> ```WikiPreview``` and moved it to **_wiki types_** file from **_nav-search service_** since it is been used in many places other than nav search

# Links
https://monopedia-ui-git-wiki-preview-on-link-hover-prediqt.vercel.app/wiki/long-sample-wiki

# Screenshot
screely-1650037562214

screely-1650037489928


src/components/Wiki/WikiPage/WikiMainContent.tsx
```diff
@@ -7,9 +7,14 @@ import { HeadingProps } from 'react-markdown/lib/ast-to-react'
 interface WikiMainContentProps {
   wiki: Wiki | undefined
   addToTOC: (props: React.PropsWithChildren<HeadingProps>) => JSX.Element
+  addWikiPreview: any
```
kesar, line 10: can we use JSX.Element instead of any? and in the function definition?

Royal-lobster: Ohh yes. i was searching for the props type but unable to find for Links. i will try make a new type for that

kesar (approved):

PR #174, opened 2022-04-18
History Page for Wiki

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/220

# Links
https://monopedia-ui-git-history-page-prediqt.vercel.app/wiki/sushiswap/history

# Screenshots 
screely-1650297681051

# Tasks
- [x] figure out which route the history page should reside
- [x] add activity card as latest wiki card at the top of history page
- [x] update the link in action bar in wiki page
- [x] implement tree structure for wiki cards
- [x] make tree structure for wiki cards mobile responsive
- [x] implement wiki history card
- [x] integrate data with history page  
- [x] Think of a way to get what changed information
- [x] Integrate Edit details with history page


src/components/Wiki/History/HistoryCard.tsx
```diff
@@ -131,12 +154,12 @@ export const HistoryCard = ({
...
           </Text>
           <Link
-            href="https://etherscan.io/tx/0x1234567890"
+            href={`https://etherscan.io/tx/${transactionAddress}`}
             color="brand.500"
             ml={2}
             isExternal
           >
-            {shortenAccount('0x1234567890')}
+            {shortenAccount(transactionAddress || '')}
```
kesar, line 162: we should try to avoid the || '' that I see often in the code.

this shortenAccount should never be rendered if transactionAddress is not a string so we need to handle that on a higher level location or have a block w a condition that ensures this is not rendered if transactionAddress is not a string 

(general comment for other any other piece of the code) this OR trick is a code smell that indicates that state is not handled properly. 👍🏻

Royal-lobster: Ohh thanks i don't know that. I will fix it !

src/pages/wiki/[slug]/history.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return (
+              <HistoryCard
+                key={activity.id}
+                isRightAligned={isHistoryFullWidth ? true : index % 2 === 0}
+                isFullWidth={isHistoryFullWidth}
+                lastEditor={activity.content[0].user.id}
+                lastEditedTime={activity.datetime}
+                transactionAddress={activity.content[0].transactionHash}
+                commitMessage={
+                  getActivityMetadataById(activity, 'commit-message')?.value
```
Comment (kesar) [inline src/pages/wiki/[slug]/history.tsx:84]: all these strings can probably get from the enum that you might implement in your other PR

kesar (approved):

src/pages/wiki/[slug]/history.tsx
Comment (Royal-lobster) [reply src/pages/wiki/[slug]/history.tsx:84]: ohh yes that will be great ! i will add commit them it that PR

src/components/Wiki/History/HistoryCard.tsx
```diff
@@ -0,0 +1,219 @@
...
+              _dark={{ boxShadow: '0 0 0 1px rgba(255, 255, 255, 0.16)' }}
+              color={{ light: 'black', _dark: 'white' }}
+              size="sm"
+              key={i}
+              gap={1}
+            >
+              <Text>{changed}</Text>
+
+              {changed === 'content' && (
+                <Tooltip label={`${wordsChanged} words changed`}>
```
kesar, line 187: for this and all other metadata numbers, it would be great to have some small validation on the UI to make sure we get a number (or a string of a number). 

Ideally we will curate it in our backend, but just to avoid someone hijacking the page bypassing backend validation

Royal-lobster: Number in the sense of value ? will that be ok ? i think it may become slow if content becomes large and there are many history cards to calculate since the diff algorithm is based on the content length.

Royal-lobster: Ohh got it ! i will make that check

kesar (approved):

PR #175, opened 2022-04-18
Fixing Avatar on Hero and ENS domain

codiejay:
Now the avatar only appears when it is fetched completely. I had to make the placeholder avatars optional. In this case if a placeholder avatar appears it could make the user try to click it to go to the user's profile but that will take them to the wiki article instead. So now the avatar will only appear when it is returned. 

Updates: An issue i just noticed is that links wouldn't work on the <CustomAvatar /> component and that's why we have the regular placeholder when a user does not have an avatar attached to their account. This way, when clicked it takes to the account page. 

## Linked issues

closes #233, closes #233, closes https://github.com/EveripediaNetwork/issues/issues/233


Royal-lobster (changes requested):
I think the layout near the avatar is [shifting](https://web.dev/cls/) too much which might be sudden for users, i think it looks good if we add a shiny [skeleton](https://chakra-ui.com/docs/components/feedback/skeleton) for avatar while loading so it fixes that.  

Maybe we can also implement skeletons with a circle and two stacked horizontal rectangles at the side for the avatar, wiki name and username instead of loading with three dots 

image

what do you think?

Royal-lobster (commented):
There seems to be a problem with image, the placeholder is not getting updated with avatar 

image

codiejay:
> There seems to be a problem with image, the placeholder is not getting updated with avatar
> 
> image

@Royal-lobster oh this is because this last user to edit this wiki doesn’t have an avatar. In cases like this I should use the CustomAvatar component that displays a better looking placeholder avatar but when I do that the link to the users account page no longer works.

Royal-lobster:
> @Royal-lobster oh this is because this last user to edit this wiki doesn’t have an avatar. In cases like this I should use the CustomAvatar component that displays a better looking placeholder avatar but when I do that the link to the users account page no longer works. 

Ohh that's weird. did you find why the link is not working on custom avatar? I am wondering what may be causing the link to break. maybe we can link the user profile inside of  ```Avatar``` by default if that works since it is the logical behaviour.

codiejay:
> > @Royal-lobster oh this is because this last user to edit this wiki doesn’t have an avatar. In cases like this I should use the CustomAvatar component that displays a better looking placeholder avatar but when I do that the link to the users account page no longer works.
> 
> Ohh that's weird. did you find why the link is not working on custom avatar? I am wondering what may be causing the link to break. maybe we can link the user profile inside of `Avatar` by default if that works since it is the logical behaviour.

I have solved this locally now. Thanks for the suggestion. I'll push hopefully soon.

src/components/Landing/Hero.tsx
```diff
@@ -35,17 +36,20 @@ const HeroCard = ({ wiki }: HeroProps) => {
...
             justify="space-between"
             fontWeight="semibold"
           >
             <chakra.span>{wiki?.title}</chakra.span>
-            <chakra.span color="blue">
-              {shortenAccount(wiki?.user?.id || '')}
-            </chakra.span>
+
+            <NextLink href={`/account/${wiki?.user?.id}`} passHref>
+              <chakra.span color="blue">
```
anubra266, line 49: FOr accessibility issues, the child receiving `passHref` should be an `a` tag. 
```suggestion
            <NextLink href={`/account/${wiki?.user?.id}`} passHref>
              <chakra.a color="blue">
```

src/components/Profile/UserDetails.tsx
```diff
@@ -29,6 +29,7 @@ export const UserDetails = (props: UserDetailsProps) => {
 
   const { headerIsSticky } = useProfileContext()
   const [, username, loading] = useENSData(address)
+  console.log(username)
```
anubra266, line 32: Can we remove this?

Royal-lobster (approved):

PR #180, opened 2022-04-19
Add tags to navbar search

anubra266:
Resolves EveripediaNetwork/issues#256

### DIFF
--- a/src/services/nav-search/utils.ts
+++ b/src/services/nav-search/utils.ts
@@ -1,4 +1,5 @@
 import { getCategoriesByTitle, getWikisByTitle } from '@/services/nav-search'
+import { getTagWikis } from '@/services/wikis'
 import { store } from '@/store/store'
 import { Category } from '@/types/CategoryDataTypes'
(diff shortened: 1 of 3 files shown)


src/services/nav-search/utils.ts
```diff
@@ -56,6 +58,7 @@ export const useNavSearch = () => {
     if (query && query.length >= 3) {
       setIsLoading(true)
       debouncedFetchResults(query, res => {
+        console.log('res.articles :>> ', res.articles)
```
kesar, line 61:
```suggestion
```

kesar:
i think tags should be float right based on design:


Captura de pantalla 2022-04-20 a las 9 29 45
Captura de pantalla 2022-04-20 a las 9 30 51

anubra266:
> i think tags should be float right based on design:
> 
> Captura de pantalla 2022-04-20 a las 9 29 45
> 
> Captura de pantalla 2022-04-20 a las 9 30 51

Fixed

kesar (approved):

PR #181, opened 2022-04-20
Infinite scroll on tags page

Emmanueldmlr:
# Implementation of Infinite scroll on tag's page

_Instead of fetching wikis that have a particular tag at once, we could fetch them in chunks as the user scrolls down the page. By so doing, the performance of the page is much improved.

## How should this be tested?

1. Navigate to https://alpha.everipedia.org/tags/hello
2. scroll down on the page, wikis with the tag "hello" are fetched in chunks as you scroll down 

image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/260

### DIFF
--- a/src/pages/tags/[tag].tsx
+++ b/src/pages/tags/[tag].tsx
@@ -1,4 +1,4 @@
-import React, {useState, useEffect} from 'react'
+import React, { useState, useEffect } from 'react'
 import { NextPage, GetServerSideProps } from 'next'
 import {
   Divider,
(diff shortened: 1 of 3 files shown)


kesar:
if tags can have footer, can you add back footer to categories? I know its unrelated to this PR but it would be great to keep cohesion across the site 👍🏻

kesar (changes requested):
if tags can have footer, can you add back footer to categories? I know its unrelated to this PR but it would be great to keep cohesion across the site 👍🏻

Emmanueldmlr:
Okay then, I will do that

kesar (approved):

PR #183, opened 2022-04-20
Preserve Newlines like Editor

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/273

screely-1650466404396

### DIFF
--- a/src/components/Wiki/WikiPage/WikiMainContent.tsx
+++ b/src/components/Wiki/WikiPage/WikiMainContent.tsx
@@ -51,7 +51,7 @@ const WikiMainContent = ({
             colorMode === 'dark' ? 'markdown-body-dark' : ''
           }`}
         >
-          {wiki?.content || ''}
+          {wiki?.content.replace(/\n/gm, '  \n') || ''}
         </ReactMarkdown>
       </Box>
     </Box>
--- a/src/components/Wiki/WikiPage/WikiMainContent.tsx
+++ b/src/components/Wiki/WikiPage/WikiMainContent.tsx
@@ -51,7 +51,7 @@ const WikiMainContent = ({
             colorMode === 'dark' ? 'markdown-body-dark' : ''
           }`}
         >
-          {wiki?.content.replace(/\n/gm, '  \n') || ''}
+          {wiki?.content || ''}
         </ReactMarkdown>
       </Box>
     </Box>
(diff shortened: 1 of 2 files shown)


Royal-lobster:
Ohh yes, it seems the editor is fine handling it. i will make it do the conversion on save !

Royal-lobster:
Opps i think it's having problem with WYSIWYG editor, the line breaks are disappearing

Royal-lobster:
Yes there seems to be problem with wysiwyg editor and line breaks. i tried to again replace the ```(two spaces) \n``` with ```\n``` while its is setting the initial state but wysiwyg editor is not working. it looks like the sync is happening between the editors internally https://github.com/nhn/tui.editor/issues/485#issuecomment-679468247 they seems to be updating that editor so we can have a native functionality in future

kesar:
wow, what if we do a conversion on save and a "revert" on wiki edit so we do the replacements only in wiki editor but not in wiki view that will be visited by many more users

Royal-lobster:
> wow, what if we do a conversion on save and a "revert" on wiki edit so we do the replacements only in wiki editor but not in wiki view that will be visited by many more users

Nope, tried that but it didn't work. it is even not working in main. i think its a bug of its own. 

once check out this page 
https://alpha.everipedia.org/create-wiki?slug=gartic-famous-unknown-answer 

i did not save it with line breaks so it's fresh from main. when we click on wysiwyg tab at the bottom, the soft line breaks again go back to single line and pressing the markdown tab again, the content remains in single line hence losing the line breaks. 

but surprisingly its working fine when we write normally on editor in markdown and switching to wysiwyg tag. i guess it's maintaining some state sync. I am not sure if we can replicate it.

Royal-lobster:
Many are having the same issue it seems - https://github.com/nhn/tui.editor/issues/485 aparently the editor is not following the proper GFM specs for soft line break.  

https://user-images.githubusercontent.com/52039218/164283159-979a715e-659b-4290-997a-f8801b6902b1.mov

I guess we can merge this for now as it works as same as before with soft line breaks feature.  

the problem remains the same when the user switches the editor to wysiwyg. if there is soft line breaks in content, then they wrap to single line again. we can address this in further iteration

kesar (approved):

Royal-lobster:
Finally i fixed it ! I was trying to only set the state by dispatch and forgot about setting the editor value 😭😭😭 now it is working as expected.

https://user-images.githubusercontent.com/52039218/164297478-5f822e3e-2c53-4be2-bd19-aea6b588711e.mov

Now it formats the softbreaks with spaces while publishing, and again revert back to no spaces while loading the editor initial state as you suggested

PR #185, opened 2022-04-21
Add uniswap widget

anubra266:
Resolves EveripediaNetwork/issues#248

CleanShot 2022-04-21 at 09 28 23@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@rtk-query/graphql-request-base-query": "^1.0.3",
     "@toast-ui/editor": "^3.1.3",
     "@toast-ui/react-editor": "^3.1.3",
+    "@uniswap/widgets": "^1.0.0",
     "axios": "^0.26.1",
     "bigpicture": "^2.6.1",
     "boring-avatars": "^1.6.3",
--- a/src/components/Wiki/WikiPage/InsightComponents/Uniswap.tsx
+++ b/src/components/Wiki/WikiPage/InsightComponents/Uniswap.tsx
@@ -1,0 +1,27 @@
+import React from 'react'
+
+import { SwapWidget, darkTheme, lightTheme } from '@uniswap/widgets'
+// import '@uniswap/widgets/dist/fonts.css'
+
+import { useColorModeValue } from '@chakra-ui/color-mode'
+import { provider } from '@/pages/_app'
+
+export type UniswapProps = {}
+// Infura endpoint
+const jsonRpcEndpoint = process.env.NEXT_PUBLIC_ENS_RPC
+
+export const Uniswap = (props: UniswapProps) => {
+  const {} = props
+  const theme = useColorModeValue(lightTheme, darkTheme)
+  const uniswapProvider = provider()
+
+  return (
+    <div>
+      <SwapWidget
+        provider={uniswapProvider}
+        theme={theme}
+        jsonRpcEndpoint={jsonRpcEndpoint}
+      />
+    </div>
+  )
+}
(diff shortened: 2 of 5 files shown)


kesar:
really promising! two things

could we make border radius a bit more similar to the ones that we have in the app, that way it will look even more integrated.


somehow we need to be able to reuse the connection we already have w our login system, I logged in w my metamask account but widget is not catching it, so I am not able to use it. That's probably the more complex part, but without that, widget wont be operative.

kesar:
i think its the provider: https://github.com/EveripediaNetwork/ep-ui/pull/185/files#diff-2a119dc424cd23f948ec46d44036cd70a66a64bbfcfcdebecc7dae53f5621c40R19

instead of uniswap provider, i think you need to use the wagmi provider: 

https://wagmi.sh/docs/hooks/useProvider

anubra266:
> i think its the provider: https://github.com/EveripediaNetwork/ep-ui/pull/185/files#diff-2a119dc424cd23f948ec46d44036cd70a66a64bbfcfcdebecc7dae53f5621c40R19
> 
> instead of uniswap provider, i think you need to use the wagmi provider:
> 
> https://wagmi.sh/docs/hooks/useProvider

@kesar I'm actually using wagmi provider, I'm importing it from `_app.tsx`. I only gave it a variable name of `uniswapProvider`

kesar:
you need to use useProvider to get the current hooked one I think that its in the context

anubra266:
> you need to use useProvider to get the current hooked one I think that its in the context

@kesar just tried it. Didn't change anything. 😢

kesar:
perhaps you need to re-render component once provider change / its logged in.

another option it could be use something like https://wagmi.sh/docs/hooks/useConnect and wait till its connected to re-render / pass the provider again.

if none of those work, they have a discord channel for support: https://discord.com/channels/597638925346930701/941447445844463676


as this is something new, we could be hitting some bug

anubra266:
> perhaps you need to re-render component once provider change / its logged in.
> 
> another option it could be use something like https://wagmi.sh/docs/hooks/useConnect and wait till its connected to re-render / pass the provider again.
> 
> if none of those work, they have a discord channel for support: https://discord.com/channels/597638925346930701/941447445844463676
> 
> as this is something new, we could be hitting some bug

@kesar found this in their docs
CleanShot 2022-04-21 at 10 10 09@2x

I'll look for help in their discord though

kesar:
> > perhaps you need to re-render component once provider change / its logged in.
> > another option it could be use something like https://wagmi.sh/docs/hooks/useConnect and wait till its connected to re-render / pass the provider again.
> > if none of those work, they have a discord channel for support: https://discord.com/channels/597638925346930701/941447445844463676
> > as this is something new, we could be hitting some bug
> 
> @kesar found this in their docs CleanShot 2022-04-21 at 10 10 09@2x
> 
> I'll look for help in their discord though

mmm then it could be that we are using matic network on the site but in the rpc we are pointing to ethereum network

anubra266:
@kesar that discord link didn’t help though. 😁 
CleanShot 2022-04-21 at 10 15 39@2x

kesar:
> @kesar that discord link didn’t help though. 😁 CleanShot 2022-04-21 at 10 15 39@2x

https://discord.com/invite/FCfyBSbCU5

they have then a wrong link in the docs :D

anubra266:
@kesar I now get this.
CleanShot 2022-04-21 at 11 15 41@2x

But the wallet seems to be conected

PR #187, opened 2022-04-21
Edit Stats in Wiki Metadata

Royal-lobster:
# Changes
Calculates what changed information in an edit session of a wiki in create-wiki page to supplement history pages which contains -
1. Percent content changed
2. Content changed word count
3. Blocks changed in wiki like title, summary, metadata etc

# Screenshots
image
 
# Tasks
- [x] Refactor metadata state for edit stats content
- [x] Fix commit message not starting empty when wiki edit page is initialized
- [x] Calculate percent content changed and update state
- [x] Calculate words changed and update state
- [x] Calculate which blocks have changed and update state
- [x] Make ```calculateEditInfo``` to use previous wiki from previous edit in same session if there is any


kesar (approved):

src/pages/create-wiki/index.tsx
```diff
@@ -227,34 +235,151 @@ const CreateWiki = () => {
 
     return true
   }
+  const calculateEditInfo = (prevWiki: Wiki, currWiki: Wiki) => {
```
kesar, line 238: a page usually should be as slim as possible. Components should take their own responsibilities and use util functions to make components also small.

in this case create wiki page is > 600 loc.

not need to refactor but take in consideration this for future PR / refactorings 👍🏻

src/pages/create-wiki/index.tsx
```diff
@@ -227,34 +235,151 @@ const CreateWiki = () => {
...
+
+      const percentChanged =
+        ((contentAdded + contentRemoved) / contentUnchanged) * 100
+      const wordsChanged = wordsAdded + wordsRemoved
+
+      // update metadata in redux state
+      dispatch({
+        type: 'wiki/updateMetadata',
+        payload: {
+          id: 'words-changed',
```
kesar, line 274: you can convert editSpecificMetaIds into an enum or a different structure and use it in dispatch instead of putting strings that can lead to unexpected bugs / typos

kesar (approved):

PR #190, opened 2022-04-26
Improve buttons on wiki process dialog box

Emmanueldmlr:
# Improve buttons on the wiki process dialog box

_Once JSON is sent to our API button for view, IPFS should be active (once we get the Ipfs hash returned from API)
view on the blockchain should only be clickable when it's in the last step and we have TxId

## How should this be tested?

1. Navigate to the create-wiki page
2. Try to create a wiki
3. On the wiki process dialog box, the Ipfs button is active when the Ipfs hash is returned and the block explorer button is clickable when it's in the last step and we have TxId

## Linked issues
fixes https://github.com/EveripediaNetwork/issues/issues/274

### DIFF
--- a/src/components/Elements/Modal/WikiProcessModal.tsx
+++ b/src/components/Elements/Modal/WikiProcessModal.tsx
@@ -127,9 +127,7 @@ const WikiProcessModal = ({
                   }
                   variant="link"
                   fontWeight="semibold"
-                  size="xs"
-                  _hover={{ textDecoration: 'none' }}
-                  color="link"
+                  fontSize='xs'
                 >
                   View on Block Explorer
                 </Button>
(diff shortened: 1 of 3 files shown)


src/components/Elements/Modal/WikiProcessModal.tsx
```diff
@@ -47,7 +47,7 @@ const WikiProcessModal = ({
   const cancelRef = React.useRef<FocusableElement>(null)
   const router = useRouter()
   const handleBlockExplorer = () => {
-    if (activeStep === 3) {
+    if (activeStep === 3 && txHash) {
```
kesar, line 50: this is preventing click, but shouldnt button be disabled until its clickable?

kesar (approved):

PR #192, opened 2022-04-27
Fix/wiki card layou

anubra266:
Resolves https://github.com/EveripediaNetwork/issues/issues/289
CleanShot 2022-04-27 at 19 28 57@2x
CleanShot 2022-04-27 at 19 29 19@2x


kesar:
@Simplypheyie96 can you review it and approve / request changes since you created the issue? thanks!

Simplypheyie96 (commented):
@Emmanueldmlr looking really good. 
sorry. I just noticed something when I was reviewing that didn't cross my mind before. 

1. this guy, we already have about page on the nav bar under resources and also in the footer, I don't think the learn more is necessary again. what do you think @kesar 
image

2. I don't the live button on these wikis in trending wikis are necessary either. They aren't serving any purpose.
image

Simplypheyie96 (approved):

PR #193, opened 2022-04-27
Allow account access using ens name

Emmanueldmlr:
# Allow Account Access Using ENS Name

_Page should detect if address starts by 0x or by a different thing, and try to reverse from ENS address to their real eth address.

## How should this be tested?

1. Navigate to /account/kesar.eth
2. The page should be able to reverse the ENS name to its address and fetch user's wiki based on the  address.

Before 
image

After
image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/212

### DIFF
--- a/src/hooks/useENSData.ts
+++ b/src/hooks/useENSData.ts
@@ -4,6 +4,7 @@ import { AvatarResolver } from '@ensdomains/ens-avatar'
 import config from '@/config'
 import { useAppDispatch, useAppSelector } from '@/store/hook'
 import { addENSAddress } from '@/store/slices/ens-slice'
+import { validateAddress } from '@/utils/validateAddress'
 
 export const useENSData = (address: string | undefined | null) => {
   const [avatar, setAvatar] = useState<string>()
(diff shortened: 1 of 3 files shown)


src/hooks/useENSData.ts
```diff
@@ -14,8 +15,13 @@ export const useENSData = (address: string | undefined | null) => {
 
   useEffect(() => {
     const getAvatar = async (addrs: string) => {
+      let lookupAddress = addrs
       const provider: BaseProvider = new StaticJsonRpcProvider(config.ensRPC)
-      const name = await provider.lookupAddress(addrs)
+      if (!validateAddress(addrs)) {
```
kesar, line 20: you shouldnt send an invalid address to useENSData and validate it inside and do some tweaks. That adds unnecessary complexity into this hook. 

it also does a resolve name and then lookup again, which its doing things twice.

I would suggest that you check if its an ENS domain before reaching this point and if so, then you do the resolveName. I think you can simplify all this code doing it at the page level, and doing a redirect to the address profile.

so whenever hit the account page, check the address and if address its not 0x... try a resolve and redirect to the account w the address 👍🏻

Emmanueldmlr:
@Simplypheyie96  kindly review the 404 error page and @kesar, the update on the Ens name.

kesar (approved):

PR #194, opened 2022-04-28
Updates wiki category title

Emmanueldmlr:
# Minor Improvement on the Landing page

_Changes category title, section margin and removes all categories card

## How should this be tested?

1. Navigate to the homepage
2. scroll to the category section

Before
image

After
image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/290

### DIFF
--- a/src/components/Landing/CategoriesList.tsx
+++ b/src/components/Landing/CategoriesList.tsx
@@ -1,5 +1,12 @@
 import React, { useEffect } from 'react'
-import { LinkBox, LinkOverlay, Text, SimpleGrid, Spinner, Center } from '@chakra-ui/react'
+import {
+  LinkBox,
+  LinkOverlay,
+  Text,
+  SimpleGrid,
+  Spinner,
+  Center,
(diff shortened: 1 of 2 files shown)


kesar:
if its something design related by mabel issues, plz add her as reviewer so she can check that  👍🏻

Simplypheyie96:
@Emmanueldmlr looking really good. 
sorry. I just noticed something when I was reviewing that didn't cross my mind before. 

1. this guy, we already have about page on the nav bar under resources and also in the footer, I don't think the learn more is necessary again. what do you think @kesar 
image

2. I don't the live button on these wikis are necessary either. They aren't serving any purpose.
image

Simplypheyie96 (changes requested):
Remove 'Live' buttons in wikis under trending wiki @Emmanueldmlr

Emmanueldmlr:
Thanks for the suggestions.

Emmanueldmlr:
@Simplypheyie96, can you confirm everything is cool now?.

PR #251, opened 2022-05-09
Fix not possible to upload images

Emmanueldmlr:
# Fix issue relating to Image Upload when creating a wiki

_When creating a wiki, there is a need to upload an image. errors relating to Image upload were not properly handled, as a result, the wiki upload progress dialog keeps uploading without showing information.

## Notes or observations
To fix this issue, I tried to wrap the Image upload function with try and catch so as to handle any error that might come up. and inside the create wiki component, I'm checking if there is a value for the Ipfs before proceeding with the other process.

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/304
fixes https://github.com/EveripediaNetwork/issues/issues/309

### DIFF
--- a/src/components/Elements/Modal/WikiProcessModal.tsx
+++ b/src/components/Elements/Modal/WikiProcessModal.tsx
@@ -55,6 +55,7 @@ const WikiProcessModal = ({
       isOpen={isOpen}
       isCentered
       size="2xl"
+      closeOnOverlayClick={false}
     >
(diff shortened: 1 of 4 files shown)


Emmanueldmlr:
@kesar, this PR also fixes issue 309 ( dont allow to close tx dialog clicking in bg )

kesar (approved):

PR #265, opened 2022-05-17
Some semantic fixes causing react hydration errors

anubra266:
Fixed some semantic errors found while investigsting the hydration errors.


kesar:
ups, I just realized it wasnt fully solved! sorry for the mess

PR #275, opened 2022-05-20
Transaction ID to wiki pages

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/352

image

### DIFF
--- a/src/components/Wiki/WikiPage/InsightComponents/WikiDetails.tsx
+++ b/src/components/Wiki/WikiPage/InsightComponents/WikiDetails.tsx
@@ -13,18 +13,21 @@ import {
 } from '@chakra-ui/react'
 import shortenAccount from '@/utils/shortenAccount'
 import { SiIpfs } from 'react-icons/si'
+import { GoLink } from 'react-icons/go'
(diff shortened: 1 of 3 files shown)


kesar (approved):

src/components/Wiki/WikiPage/InsightComponents/WikiDetails.tsx
```diff
@@ -13,25 +13,29 @@ import {
...
   ipfsHash,
+  txHash,
   lastEditor,
   imgSrc,
 }: {
   wikiTitle: WikiPreview
   categories: BaseCategory[]
   lastEdited: string | undefined
   ipfsHash: string | undefined
+  txHash: string | undefined
```
kesar, line 38: someday we will need to refactor all those undefined and expect them to load the module. I dont see any reason why this should be loaded without expecting those params defined (edited, hash, etc)

Royal-lobster: Yes. I guess we we can remove those undefined since the api evolved now to returns them for all wikis. probably there are more of these so might require an other pr

PR #276, opened 2022-05-20
Syned markdown - editor styles

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/350

### DIFF
--- a/src/pages/static/assets/global.css
+++ b/src/pages/static/assets/global.css
@@ -70,8 +70,8 @@ html {
   margin: 0.6em 0 !important;
   font-size: 1.7em !important;
 }
-.toastui-editor-contents h4 {
-  font-size: 16px !important;
+:where(.toastui-editor-contents h4) {
+  font-size: 16px;
 }
 .toastui-editor-contents li {
   list-style-type: unset !important;
--- a/src/pages/static/assets/global.css
+++ b/src/pages/static/assets/global.css
@@ -51,18 +51,52 @@ html {
 }
 
 /* EDITOR CONTENT STYLES */
-.toastui-editor-main * {
+.toastui-editor-contents * {
   font-family: poppins, sans-serif !important;
 }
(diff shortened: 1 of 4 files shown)


src/pages/static/assets/global.css
```diff
@@ -51,18 +51,52 @@ html {
...
 }
-.toastui-editor-main h1,
-.toastui-editor-main h2,
-.toastui-editor-main h3 {
+.toastui-editor-contents h1,
+.toastui-editor-contents h2,
+.toastui-editor-contents h3,
+.toastui-editor-contents h4,
+.toastui-editor-contents h5,
+.toastui-editor-contents h6 {
```
anubra266, line 62: - You can shorten this kind of code this way
```suggestion
.toastui-editor-contents :where(h1, h2, h3, h4, h5, h6) {
```
- I think the where selector has higher specifity, so whenever you use it, you won't need to use `!important` test that though.

src/pages/static/assets/global.css
```diff
@@ -51,18 +51,52 @@ html {
...
+.toastui-editor-contents h2 {
+  font-weight: 500 !important;
+}
+.toastui-editor-contents h1 {
+  margin: 0.6em 0 !important;
+  font-size: 1.7em !important;
+}
+.toastui-editor-contents h4 {
+  font-size: 16px !important;
+}
```
anubra266, line 80: Just as an example to buttress my previous point
```suggestion
:where(.toastui-editor-contents h4) {
  font-size: 16px;
}
```
Note that I removed the `!important` This would help in places where you have to sacrifice theme tokens for the sake of using `!important`

src/pages/static/assets/global.css
Royal-lobster: This is awesome ! thanks

kesar (approved):

PR #280, opened 2022-05-23
Align avatar

Emmanueldmlr:
# Avatar should be near the address not in the middle vertical align

Before
image

After
image

fixes https://github.com/EveripediaNetwork/issues/issues/354

### DIFF
--- a/src/components/Landing/Hero.tsx
+++ b/src/components/Landing/Hero.tsx
@@ -10,6 +10,7 @@ import {
   VStack,
   Stack,
   Link,
+  SimpleGrid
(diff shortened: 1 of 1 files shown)


kesar:
image

I think there is no need to align right the title and subtitle and leave that empty space there. also make sure sizes are similar, avatar looks pretty big compared w text

image

Emmanueldmlr:
@kesar, what you mean?

image

kesar:
weird, it was showing different, but it looks great now 👍🏻

kesar (approved):

PR #283, opened 2022-05-23
fix new rehydration errors

anubra266:
There's a new rehydration error on the home page, I think the team should watch out and avoid introducing new rehydration errors. We're building an SSR I suppose, but these errors defeat that, and just makes the app Client side rendered.


anubra266:
@kesar I kinda got convinced through a thread that, we shouldn't use arrow functions for `useEffect` because they should explicitly say what that useEffect is there to achieve, and Dan Abramov said the new react docs will enforce that in their examples, so I disabled the eslint rule that was against this.

kesar (approved):

PR #284, opened 2022-05-23
Fix/modal closing error

anubra266:
Resolves EveripediaNetwork/issues#338

PR #287, opened 2022-05-24
Edit dialog fixes

anubra266:
Resolves EveripediaNetwork/issues#358

### DIFF
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/HighlightsModal.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/HighlightsModal.tsx
@@ -184,7 +184,12 @@ const HighlightsModal = ({
                       })
                     }
                   }}
-                  placeholder="Select Category"
+                  value={currentWiki.categories[0].title}
+                  placeholder={
+                    currentWiki.categories.length > 0
+                      ? undefined
+                      : 'Select Category'
+                  }
                 >
                   {categoryOptions?.map(o => (
                     <option key={o.title}>{o.title}</option>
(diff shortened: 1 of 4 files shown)


src/components/Layout/Editor/Highlights/HighlightsModal/HighlightsModal.tsx
```diff
@@ -184,7 +184,12 @@ const HighlightsModal = ({
                       })
                     }
                   }}
-                  placeholder="Select Category"
+                  value={currentWiki.categories[0].title}
```
kesar, line 187: isn't this assuming that there is gonna be always categories[0].title ¿?

anubra266: Yeah...that's the format it's being saved in. I think previously it allowed saving multiple categories, that's why it gets saved as an array, difference is now, the array will only contain one element

kesar: image

but you are assuming wiki has a category, if its a brand new wiki you dont have yet category assigned

anubra266: Oh...I relied on the typings, this means the typings are wrong.

anubra266, line 187:
```suggestion
                  value={currentWiki.categories[0]?.title}
```

anubra266: @kesar For a new wiki, categiories is an empty array, so I've changed it to `[0]?.title`.

kesar (approved):

Royal-lobster:
@anubra266 This problem also exists for this dialog, it is blocking the screen similar to wiki dialog

CleanShot 2022-05-25 at 20 09 00@2x

anubra266:
@Royal-lobster This PR is only for the highlights modal. Sorry if the name was confusing. You could just display the modal component when `isOpen` is true. See https://github.com/EveripediaNetwork/ep-ui/pull/284

Royal-lobster:
> @Royal-lobster This PR is only for the highlights modal. Sorry if the name was confusing. You could just display the modal component when `isOpen` is true. See #284

Ohhh my bad got it ! 😅

PR #289, opened 2022-05-25
Wagmi 0.4

kesar:
fixes https://github.com/EveripediaNetwork/issues/issues/360
fixes https://github.com/EveripediaNetwork/issues/issues/366

Test:
 ens names / avatars, wallet login metamask & magic, create wiki, edit wiki, token balances


package.json
```diff
@@ -51,7 +51,7 @@
     "chakra-ui-steps": "^1.7.2",
     "date-fns": "^2.28.0",
     "debounce": "^1.2.1",
-    "ethers": "5.6.2",
+    "ethers": "^5.6.8",
```
kesar, line 54: upgraded ethers for vulnerability https://twitter.com/ethersproject/status/1529232578670120964

src/components/Profile/SidebarFilter/ChainsFilter.tsx
```diff
@@ -19,15 +19,15 @@ import React from 'react'
 const CHAINS = [
   {
     name: 'Ethereum',
-    image: 'https://opensea.io/static/images/logos/ethereum.png',
+    image: 'https://opensea.io/static/images/logos/ethereum.svg', // TODO: store this images
```
kesar, line 22: those images were 404, but anyhow i think we should remove this piece in the future

src/config/wagmi.ts
```diff
@@ -0,0 +1,55 @@
...
+    publicProvider({ weight: 3 }),
+  ],
+)
+
+const network: Network = {
+  name: config.alchemyChain,
+  chainId: Number(config.chainId),
+}
+
+export const connectors: Connector[] = [
```
kesar, line 31: deprecated coinbase that uses not many people and makes UX more complex
switched from injected to metamask connector

src/config/wagmi.ts
```diff
@@ -0,0 +1,55 @@
...
+import { publicProvider } from 'wagmi/providers/public'
+import { AlchemyProvider, Network } from '@ethersproject/providers'
+import config from './index'
+
+type Connector =
+  | MetaMaskConnector
+  | WalletConnectConnector
+  | MagicConnector
+
+export const { chains, provider } = configureChains(
```
kesar, line 17: now it checks chains mumbai & polygon instead of all.
i also set weights to providers will be tested in the order provided, giving more weight to alchemy but w possibility to failback in others if alchemy is down

src/config/wagmi.ts
Emmanueldmlr: Nice, I noticed you already removed it. Since many people don't use it, It shouldn't be an issue then.

Emmanueldmlr (approved):

PR #297, opened 2022-05-28
Improve slug generator

codiejay:
In this PR we now have the "slugifyer" inside of the utils folder. This PR has also added the `strict` and `lower` attribute as `true` when slugifying data. 

closes #371, closes #371, closes https://github.com/EveripediaNetwork/issues/issues/371

### DIFF
--- a/src/pages/create-wiki/index.tsx
+++ b/src/pages/create-wiki/index.tsx
@@ -149,7 +149,7 @@ const CreateWikiContent = () => {
   const getImageHash = async () =>
     isWikiBeingEdited ? ipfsHash : saveImage(image)
 
-  const getWikiSlug = () => slugifyText(String(wiki.title).toLowerCase())
+  const getWikiSlug = () => slugifyText(String(wiki.title))
 
   const getImageArrayBufferLength = () =>
     (image.type as ArrayBuffer).byteLength === 0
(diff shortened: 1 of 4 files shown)


src/utils/slugify.ts
```diff
@@ -0,0 +1,9 @@
+import slugify from 'slugify'
+
+export const slugifyText = (string: string) => {
```
kesar, line 3: perhaps to avoid confusion we can rename var name from string to text. besides that, everything looks great 👍🏻

kesar (approved):

src/pages/create-wiki/index.tsx
```diff
@@ -149,7 +149,7 @@ const CreateWikiContent = () => {
   const getImageHash = async () =>
     isWikiBeingEdited ? ipfsHash : saveImage(image)
 
-  const getWikiSlug = () => slugify(String(wiki.title).toLowerCase())
+  const getWikiSlug = () => slugifyText(String(wiki.title).toLowerCase())
```
kesar, line 152: sorry, just saw it now.

instead of doing toLowerCase() here and in https://github.com/EveripediaNetwork/ep-ui/pull/297/files#diff-e40ce2cbc02291ef08c48d4610ca97789821152c9c0cfd5821dd208f82b55558R177

we could do lowercase inside of slugifyText that way, if we need to add it in other place we wont need to add toLowercase and it will avoid some bugs 👍🏻

codiejay: Done 🍰

PR #301, opened 2022-05-31
Remove empty metadata object

codiejay:
In this PR, before posting a newly created wiki or editing an already existing one, we make sure to filter the metadata to make sure what only passes are those with a value. But we are allowing `reference` to pass through to avoid a bug.  

closes #343, closes #343, closes https://github.com/EveripediaNetwork/issues/issues/343


src/pages/create-wiki/index.tsx
```diff
@@ -261,7 +261,9 @@ const CreateWikiContent = () => {
       // Build the wiki object after edit info has been calculated
       const finalWiki = {
         ...interWiki,
-        metadata: store.getState().wiki.metadata,
+        metadata: store.getState().wiki.metadata.filter(meta => {
+          return meta.value !== '' || meta.id === 'references'
```
Royal-lobster, line 265: the references in string can use enum 
```suggestion
          return meta.value !== '' || meta.id === CommonMetaIds.REFERENCES
```

kesar:
code looks good. have you tested it editing / creating wikis to see that it works as expected? if so, we can merge it

codiejay:
> code looks good. have you tested it editing / creating wikis to see that it works as expected? if so, we can merge it

Yeah it works good, but I am implementing the code review by @Royal-lobster. The issue now is for some reason my account is not allowing me create wikis to test with. But should push changes.

kesar (approved):

PR #306, opened 2022-06-02
Tag suggestions

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/329

# Links
- https://monopedia-ui-git-tag-suggestions-prediqt.vercel.app/create-wiki?slug=sushiswap

# Screenshots
https://user-images.githubusercontent.com/52039218/171870282-fbb3fb04-972f-45c8-b4f3-eb1393ce8730.mp4

### DIFF
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -65,6 +65,9 @@ const Tags = () => {
           if (suggestionSelectionId > 0) setSuggestionSelectionId(p => p - 1)
         },
         Enter() {
+          setQuery('')
+          setSuggestionSelectionId(-1)
+          inputRef.current?.focus()
           if (suggestionSelectionId === -1) return
           if (editTagIndex === -1)
             api.setValue([...api.value, results[suggestionSelectionId]?.id])
@@ -77,9 +80,6 @@ const Tags = () => {
               ),
             )
           }
-          setQuery('')
-          setSuggestionSelectionId(-1)
-          inputRef.current?.focus()
         },
       }
       const exec = keyMap[e.key]
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -75,7 +75,7 @@ const Tags = () => {
           sx={{ ...tagsInputStyle }}
         >
           {api.value.map((value, index) => (
-            <span key={index}>
+            <chakra.span key={index} whiteSpace="nowrap">
               <div {...api.getTagProps({ index, value })}>
                 <span>{value} </span>
                 <button
@@ -86,9 +86,11 @@ const Tags = () => {
                 </button>
               </div>
               <input {...api.getTagInputProps({ index, value })} />
-            </span>
+            </chakra.span>
           ))}
-          <input placeholder="Add tag..." {...InputProps} ref={inputRef} />
+          {api.value.length < 5 && (
+            <input placeholder="Add tag..." {...InputProps} ref={inputRef} />
+          )}
         </chakra.div>
         {tagState.invalid ? (
           <Text fontSize="xs" color="red">
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -1,11 +1,12 @@
 import React, { memo, useRef, useState } from 'react'
-import { Stack, Text, chakra, Box } from '@chakra-ui/react'
+import { Stack, Text, chakra, Box, HStack, Icon } from '@chakra-ui/react'
 import * as tagsInput from '@zag-js/tags-input'
 import { mergeProps, useMachine, useSetup } from '@zag-js/react'
 
 import { useAppDispatch, useAppSelector } from '@/store/hook'
 import { tagsInputStyle } from '@/components/Layout/Editor/Highlights/HighlightsModal/styles'
 import { useTagSearch } from '@/services/search/utils'
+import { RiLightbulbLine } from 'react-icons/ri'
 
 const MAX_LENGTH = 15
 
@@ -62,8 +63,7 @@ const Tags = () => {
       <chakra.div
         rounded="md"
         border="solid 1px"
-        borderColor="gray.300"
-        _dark={{ borderColor: 'whiteAlpha.300', bg: 'gray.700' }}
+        borderColor="borderColor"
         p={3}
         pos="relative"
       >
@@ -93,7 +93,7 @@ const Tags = () => {
           )}
         </chakra.div>
         {tagState.invalid ? (
-          <Text fontSize="xs" color="red">
+          <Text fontSize="xs" color="red.300">
             {tagState.msg}
           </Text>
         ) : null}
@@ -130,7 +130,10 @@ const Tags = () => {
                   },
                 }}
               >
-                <Text>{tag.id}</Text>
+                <HStack>
+                  <Icon as={RiLightbulbLine} fontSize={20} />
+                  <Text>{tag.id}</Text>
+                </HStack>
               </Box>
             ))}
           </Box>
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -55,7 +55,7 @@ const Tags = () => {
         border="solid 1px"
         borderColor="gray.300"
         _dark={{ borderColor: 'whiteAlpha.300', bg: 'gray.700' }}
-        p="5"
+        p={4}
       >
         <chakra.div ref={ref} {...api.rootProps} sx={{ ...tagsInputStyle }}>
           {api.value.map((value, index) => (
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -131,8 +131,8 @@ const Tags = () => {
                 }}
               >
                 <HStack>
-                  <Icon as={RiLightbulbLine} fontSize={20} />
-                  <Text>{tag.id}</Text>
+                  <Icon as={RiLightbulbLine} color="linkColor" fontSize={20} />
+                  <Text color="linkColor">{tag.id}</Text>
                 </HStack>
               </Box>
             ))}
--- a/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
+++ b/src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
@@ -1,5 +1,5 @@
 import React, { memo, useCallback, useEffect, useRef, useState } from 'react'
-import { Stack, Text, chakra, Box } from '@chakra-ui/react'
+import { Stack, Text, chakra, Box, useDisclosure } from '@chakra-ui/react'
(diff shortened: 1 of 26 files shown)


anubra266:
@Royal-lobster Wouldn't it be better if the suggestions show up as soon as we start typing

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
```diff
@@ -70,15 +86,58 @@ const Tags = () => {
                 </button>
               </div>
               <input {...api.getTagInputProps({ index, value })} />
```
anubra266, line 88: We would also want to hook into the individual tag inputs, so users get suggestions while editing existing tags.

Royal-lobster:
given the 3 characters min limit so it is not showing. do you think we should change that to no minimum limit ?

anubra266 (changes requested):

anubra266:
> given the 3 characters min limit so it is not showing. do you think we should change that to no minimum limit ?

@Royal-lobster I don't understand...where do we have that three character limit?

Royal-lobster:
😅 it is in backend, i copied the code from wiki search and left the minimum 3 characters rule there

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
Royal-lobster: Ohh that will be awesome ! i will implement it !

anubra266:
> 😅 it is in backend, i copied the code from wiki search and left the minimum 3 characters rule there

Let's remove it then, I don't think it will be nice to have three character limit for tags.

Royal-lobster:
I will try to create a new pr there to remove it so we can get suggestions instant

anubra266:
> I will try to create a new pr there to remove it so we can get suggestions instant

@Royal-lobster note that we still want the character limit in the wiki search, it's only for tags we don't want the limit.

Royal-lobster:
@anubra266 made the changes, but tab selecting the suggestions with keyboard in tag edits makes the tag go to previous state, and when selected one of the suggestion it changes back to the suggestion that we selected.

https://user-images.githubusercontent.com/52039218/171971752-f9e3b1b1-9787-4cb2-b6f5-43a0400b87d6.mp4

I tried to read the documentation around it but unable to figure out how to make the edit to stay while selecting suggestions with keyboard tabs. apart from that it works fine with mouse clicks

anubra266:
@Royal-lobster what we should do is hide the suggestions, once, we're not editing. You can check that this way
```js
const isEditingTag = state.matches("editing:tag")
```

anubra266:
I think it'll be better to hook into `getTagInputProps`, and close the edit menu once the input is blurred. `onBlur` so the previous technique won't complicate handling, suggestions while adding new input.

anubra266:
@Royal-lobster I think it'll be better to hook into `getTagInputProps`, and close the edit menu once the input is blurred. `onBlur` so the previous technique won't complicate handling, suggestions while adding new input.

Royal-lobster:
@anubra266 But closing the suggestions menu on tag input blur will make keyboard navigation completely not possible since when the user hits the tab key for selecting a suggestion, it is currently blurring the tag input and for 2nd tab it is landing on first suggestion. (pressing enter now will change the blurred tag value to suggested)

is there any way we can stop it blurring the tag input on the tab ?

anubra266:
@Royal-lobsterWhat if we gave the tag input a ref, and with that ref, keep the input focused when the suggestions list is open or being navigated. `tagInputRef.current?.focus()`

Royal-lobster:
That's brilliant !!! i think i will work, thanks i will give it a shot !

Royal-lobster:
> @Royal-lobsterWhat if we gave the tag input a ref, and with that ref, keep the input focused when the suggestions list is open or being navigated. `tagInputRef.current?.focus()`
> 
> All these might be easier with a popover, wy aren't you using a popover?

Ohh for popover you mean from chakra ? does it do the same thing ?

Royal-lobster:
But when suggestions open in popover, doesn't it auto focus automatically to popover content ? so it forces the user to deal with suggestions instead of they being an optional suggestions while they are typing ?

anubra266:
> But when suggestions open in popover, doesn't it auto focus automatically to popover content ? so it forces the user to deal with suggestions instead of they being an optional suggestions while they are typing ?

@Royal-lobster https://chakra-ui.com/docs/components/overlay/popover#focus-an-element-when-popover-opens

anubra266:
@Royal-lobster This might help:

https://user-images.githubusercontent.com/30869823/172000515-d3e91c1f-f6c8-4c4f-9214-6c00d1421376.mp4


```js
function WithPopoverAnchor() {
  const [isEditing, setIsEditing] = useBoolean()
  const [color, setColor] = React.useState('red')
  const inp = useRef()

  return (
    <Popover
      isOpen={isEditing}
      onOpen={setIsEditing.on}
      onClose={setIsEditing.off}
      initialFocusRef={inp} 
      matchWidth
    >
        <PopoverAnchor>
          <Input 
            ref={inp}
            color={color}
            onBlur={setIsEditing.off}
            onFocus={setIsEditing.on}
            defaultValue='Popover Anchor'
          />
        </PopoverAnchor>
      <PopoverContent w="full">
        <PopoverBody>
         Suggestions
        </PopoverBody>
      </PopoverContent>
    </Popover>
  )
}
```

Royal-lobster:
@anubra266 Thanks ! but i think i missed something, so we are having suggestions in popover content, but since we want them to appear when we have results so we have to manually trigger the popover so ```popoverAnchor```, (tagInput in our case) doesn't get to trigger it, also since the tagInputs can be at most 5 there are multiple triggers and we also want those inputs to be focused while we we are focussed in the suggestions so it won't go back to previous state. so i am confused how exactly it will be different with popover 😰

anubra266:
@Royal-lobster The `isOpen` can be something like `isEditing&&results.length>0`

anubra266:
@Royal-lobster I just checked github's tags UX, Tab does not navigate suggestions, tab selects the focused tag. So let's do it that way.

Royal-lobster:
> 

Ohh yes, but the focus on tagInput while focusing on suggestions ? we still have to focus it manually with ```tagInputRef.current?.focus()``` right ?

Royal-lobster:
> @Royal-lobster I just checked github's tags UX, Tab does not navigate suggestions, tab selects the focused tag. So let's do it that way.

Ohh yes ! the tab is kinda handled by the tags system so we have to hack it to work with suggestions. do you think this is the right way ?

Royal-lobster:
> @Royal-lobster I just checked github's tags UX, Tab does not navigate suggestions, tab selects the focused tag. So let's do it that way.
> 
>  CleanShot.2022-06-04.at.14.38.41.mp4

Ohhhh got it !! so may be that is the correct method since github is doing it, so users have to use mouse to select suggestions and tabs to navigate through tabs as intended

anubra266:
> > @Royal-lobster I just checked github's tags UX, Tab does not navigate suggestions, tab selects the focused tag. So let's do it that way.
> 
> Ohh yes ! the tab is kinda handled by the tags system so we have to hack it to work with suggestions. do you think this is the right way ?

Yeah, let's not use buttons for the suggestions, so it doesn't mess with the tabbing, but we might still want arrow navigation, the github tags allows that. The way I would go about that is just have a state that tracks focuses suggestion, it's not focused in the dom of course, but we give the user that illusion so they can navigate with their arrows, but the Input is the one actually focused in the dom, so all the arrow key events will be tracked by the input itself, the suggestion whose id is stored in state as focused then has the styling of something focused, when the tag input receives the `Enter` key event, we then accept the value of the currently _fake focused_  suggestion as the value of the tag.

Royal-lobster:
So i will proceed with your previous suggestion to hide the suggestions when the tagInput is blurred 👍

Royal-lobster:
> Yeah, let's not use buttons for the suggestions, so it doesn't mess with the tabbing, but we might still want arrow navigation, the github tags allows that. The way I would go about that is just have a state that tracks focused suggestion, it's not focused in the dom of course, but we give the user that illusion so they can navigate with their arrows, but the Input is the one actually focused in the dom, so all the arrow key events will be tracked by the input itself, the suggestion whose id is stored in state as focused then has the styling of something focused, when the tag input receives the `Enter` key event, we then accept the value of the currently _fake focused_ suggestion as the value of the tag.

Ohh that's great, i think it will work ! i will try to implement it.

Royal-lobster:
@anubra266 Thanks for the suggestions ! i implemented the changes

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
```diff
@@ -44,6 +47,57 @@ const Tags = () => {
   )
   const ref = useSetup({ send, id: '1' })
   const api = tagsInput.connect(state, send)
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const handleKeyPress = useCallback(
+    (e: { key: string; preventDefault: () => void }) => {
```
anubra266, line 53:
```suggestion
    (e: React.KeyboardEvent<HTMLInputElement>) => {
```

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
```diff
@@ -44,6 +47,57 @@ const Tags = () => {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [results, suggestionSelectionId, editTagIndex, setQuery],
+  )
+
+  useEffect(() => {
+    window.addEventListener('keydown', handleKeyPress)
+    return () => {
+      window.removeEventListener('keydown', handleKeyPress)
+    }
+  }, [results, handleKeyPress])
```
anubra266, line 92: We can't do this, adding the event to window, means this runs anytime those keys are pressed, regardless where we are, In the video below, the navigation is going on, even though the input is not focused.

https://user-images.githubusercontent.com/30869823/172014038-315070df-fae0-4b4c-844b-fd9375a02c8f.mp4


1. We should close the suggestions, when any of the two inputs is blurred. `onBlur`. That means we should have a `disclosure` state for the results. Results should show when `results.length>0` and when the disclosure state `isOpen` is true. So we set `isOpen` true, when any of the tag inputs is focused; `onFocus` and set to false, when they're blurred.
2. We set the `onKeyDown` event handler on the inputs and not on window. Just add to the inputs' `mergeProps`; 

```ts
  const InputProps = mergeProps(propsFromAPI, {
   //...
  onKeyDown: handleKeyPress
  })
```

anubra266 (changes requested):

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
```diff
@@ -44,6 +47,57 @@ const Tags = () => {
...
+            api.value.map((_, index) =>
+              index === editTagIndex
+                ? results[suggestionSelectionId]?.id
+                : api.value[index],
+            ),
+          )
+        setQuery('')
+        setSuggestionSelectionId(-1)
+        inputRef.current?.focus()
+      }
```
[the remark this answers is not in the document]
Royal-lobster: This is best !! thanks for the suggestion ^^

src/components/Layout/Editor/Highlights/HighlightsModal/Tags.tsx
Royal-lobster: Ohh thanks ! i will change the implementation to follow these points

anubra266 (approved):

kesar:
@anubra266 @Royal-lobster kudos to the work and comments here. best code review / discussion ive seen in a while! 🧠

PR #310, opened 2022-06-03
Show commit message in revision page

Emmanueldmlr:
# Show commit message on revision page

_if the user added a commit message prob. it would be something meaningful to show in the revision page

## How should this be tested?

1. Edit a wiki
2. Add a commit message
3. Go to the wiki page to view the commit message

image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/315

### DIFF
--- a/src/components/Wiki/WikiAccordion/AccordionWidget.tsx
+++ b/src/components/Wiki/WikiAccordion/AccordionWidget.tsx
@@ -21,10 +21,12 @@ import {
   RiShareBoxLine,
 } from 'react-icons/ri'
 import shortenAccount from '@/utils/shortenAccount'
+import DisplayAvatar from '@/components/Elements/Avatar/Avatar'
+import { useENSData } from '@/hooks/useENSData'
 
 const AccordionWidget = ({ type, title, titleTag, content }: WikiInsights) => {
   const { hasCopied, onCopy } = useClipboard(content as string)
-
+  const [, username] = useENSData((type === 'address' && typeof content === 'string') ?  content : '')
   const contentTemplate = () => {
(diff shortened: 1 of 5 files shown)


kesar (approved):

moraesux:
Looks good, same as in figma.

I just had a question, I created a new wiki and even before clicking to edit, the commit appears. was that supposed to happen?

Simplypheyie96:
image
there is too much repetition going on here. The wikis already have last edit/last editor, i don't think it should be repeated in the commit message. i am just noticing this 
@moraesux @Emmanueldmlr

kesar:
> there is too much repetition going on here. The wikis already have last edit/last editor, i don't think it should be repeated in the commit message. i am just noticing this @moraesux @Emmanueldmlr

I would suggest to remove latest editor / last edit from the revision page. It gives information from the future for a wiki that yo u want to see in a particular revision. Knowing who edit that revision and when sounds more important to me

Simplypheyie96:
> > there is too much repetition going on here. The wikis already have last edit/last editor, i don't think it should be repeated in the commit message. i am just noticing this @moraesux @Emmanueldmlr
> 
> I would suggest to remove latest editor / last edit from the revision page. It gives information from the future for a wiki that yo u want to see in a particular revision. Knowing who edit that revision and when sounds more important to me

There is something i would like to propose about this in conjunction with the user profile. I'll open and issue and tag you. Maybe that would solve it. @kesar

kesar:
cool! I started issue here: https://github.com/EveripediaNetwork/issues/issues/401 

can we move convo there and add suggestion there 🙏🏻

PR #317, opened 2022-06-07
Media upload improvement

Emmanueldmlr:
# Media upload improvement

_Redesign the media upload modal following the new design on the Figma page.

## How should this be tested?

1. go to the create wiki page
2. click on the new image or video button
3. You should have two options (Drag and drop & upload a link)  through which you can upload your media

Before
image

After
image

fixes https://github.com/EveripediaNetwork/issues/issues/390

### DIFF
--- a/src/components/Elements/Dropzone/Dropzone.tsx
+++ b/src/components/Elements/Dropzone/Dropzone.tsx
@@ -93,7 +93,7 @@ const Dropzone = ({ dropZoneActions }: DropzoneType) => {
           justifyContent="center"
           alignItems="center"
           maxH="345px"
-          minH={!showFetchedImage ? "165px": "300px"}
+          minH={!showFetchedImage ? '165px' : '300px'}
           _hover={{
             boxShadow: 'md',
             borderColor: 'brand.400',
@@ -104,7 +104,7 @@ const Dropzone = ({ dropZoneActions }: DropzoneType) => {
           {isDragActive ? (
             <Text textAlign="center">Drop the files here ...</Text>
           ) : (
-            <Box px="8" mb={!showFetchedImage ? "10": "1"}>
+            <Box px="8" mb={!showFetchedImage ? '10' : '1'}>
               <Text textAlign="center" opacity="0.5">
                 Drag and drop a <b>{textType}</b>, or click to select.
               </Text>
--- a/src/components/Elements/Dropzone/Dropzone.tsx
+++ b/src/components/Elements/Dropzone/Dropzone.tsx
@@ -93,7 +93,7 @@ const Dropzone = ({ dropZoneActions }: DropzoneType) => {
           justifyContent="center"
           alignItems="center"
           maxH="345px"
-          minH="300px"
+          minH={!showFetchedImage ? "165px": "300px"}
           _hover={{
             boxShadow: 'md',
             borderColor: 'brand.400',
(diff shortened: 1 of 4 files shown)


Emmanueldmlr:
image

Simplypheyie96 (changes requested):
Looks great. Just this tiny thing. Is it possible that we can have the image link disappear once the picture is uploaded, because they might be confused with the error message that shows up when trying to clear the link to post another. 
image

JustinCast:
> image After several images were uploaded the dialog layout got cut

Maybe we can set a max media limit

Emmanueldmlr:
> > image After several images were uploaded the dialog layout got cut
> 
> Maybe we can set a max media limit

@Simplypheyie96 and @kesar  , what should we do about this?

Emmanueldmlr:
> Looks great. Just this tiny thing. Is it possible that we can have the image link disappear once the picture is uploaded, because they might be confused with the error message that shows up when trying to clear the link to post another. image

no problem. I will fix it

Simplypheyie96:
> > > image After several images were uploaded the dialog layout got cut
> > 
> > 
> > Maybe we can set a max media limit
> 
> @Simplypheyie96 and @kesar , what should we do about this?

first off, we need to reduce the size of this box, then add beside the 10mb max size, we write something like (10mb max/10media files). but the first thing that comes to mind is to have a scroll bar on there
image

Emmanueldmlr:
@Simplypheyie96, I'm not sure if a scroll bar is a better UI. I think we should reduce the max no of uploads instead.

Simplypheyie96:
image
something like this @Emmanueldmlr

Emmanueldmlr:
@Simplypheyie96, can you update the properties on Figma, so I can work with the new properties.

Simplypheyie96:
> @Simplypheyie96, can you update the properties on Figma, so I can work with the new properties.

it is on figma

kesar:
image

👍🏻

Royal-lobster:
@Emmanueldmlr This is working awesome ! 

And i noticed that the file names can be a bit longer as it is truncating half way 

CleanShot 2022-06-08 at 14 14 19@2x

@Simplypheyie96, i think the heading local files no longer represent this since the files are uploaded to ipfs immediately but i am not sure if it may mislead some users

Emmanueldmlr:
> @Emmanueldmlr This is working awesome !
> 
> And i noticed that the file names can be a bit longer as it is truncating half way
> 
> CleanShot 2022-06-08 at 14 14 19@2x
> 
> @Simplypheyie96, i think the heading local files no longer represent this since the files are uploaded to ipfs immediately but i am not sure if it may mislead some users

I will fix it now
@Simplypheyie96, I'm still waiting for your review.

Simplypheyie96:
> > @Emmanueldmlr This is working awesome !
> > And i noticed that the file names can be a bit longer as it is truncating half way
> > CleanShot 2022-06-08 at 14 14 19@2x
> > @Simplypheyie96, i think the heading local files no longer represent this since the files are uploaded to ipfs immediately but i am not sure if it may mislead some users
> 
> I will fix it now @Simplypheyie96, I'm still waiting for your review.

It looks great. 
For the local files that @Royal-lobster suggested, it can be changed to Uploads @Emmanueldmlr 
image

Emmanueldmlr:
Just to confirm, change Local files to Upload files?... @Simplypheyie96

Simplypheyie96:
> Just to confirm, change Local files to Upload files?... @Simplypheyie96

no, to just "Uploads"... indicating to the users that files they upload will be displayed in the box below.
image

Emmanueldmlr:
image

@Simplypheyie96 @Royal-lobster , fixed all issues now.

kesar (approved):

Royal-lobster (approved):

PR #319, opened 2022-06-07
New changes to blog

anubra266:
Resolves EveripediaNetwork/issues#409


Simplypheyie96 (changes requested):
Can we add more paddings to the sides? We don't want the sentences to be too long to read.
something like this: https://everipedia.org/blog/iq-monthly-news-may-2022
But not as much padding, but something more to make reading a bit easier.
(The padding is just for the large breakpoint). it looks great on other breakpoints

Simplypheyie96 (approved):
Looks awesome. Great Job!👍🏼

kesar (approved):

PR #325, opened 2022-06-08
fixed social links in commit messages

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/412

### DIFF
--- a/src/types/Wiki.ts
+++ b/src/types/Wiki.ts
@@ -28,12 +28,12 @@ export enum EditorContentOverride {
 export enum CommonMetaIds {
   PAGE_TYPE = 'page-type',
   REFERENCES = 'references',
-  FACEBOOK_PROFILE = 'facebook-profile',
-  INSTAGRAM_PROFILE = 'instagram-profile',
-  TWITTER_PROFILE = 'twitter-profile',
-  LINKEDIN_PROFILE = 'linkedin-profile',
-  YOUTUBE_PROFILE = 'youtube-profile',
-  COINGECKO_PROFILE = 'coingecko-profile',
+  FACEBOOK_PROFILE = 'facebook_profile',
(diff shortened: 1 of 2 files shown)


src/types/Wiki.ts
```diff
@@ -28,12 +28,12 @@ export enum EditorContentOverride {
 export enum CommonMetaIds {
   PAGE_TYPE = 'page-type',
   REFERENCES = 'references',
-  FACEBOOK_PROFILE = 'facebook_profile',
-  INSTAGRAM_PROFILE = 'instagram_profile',
-  TWITTER_PROFILE = 'twitter_profile',
-  LINKEDIN_PROFILE = 'linkedin_profile',
-  YOUTUBE_PROFILE = 'youtube_profile',
-  COINGECKO_PROFILE = 'coingecko_profile',
+  FACEBOOK_PROFILE = 'facebook-profile',
```
kesar, line 31: if you change this, api needs to be aware and legacy wikis wont work

Royal-lobster: Ohh shall i change it back to underscores ? i made it hyphens to make those similar to rest

kesar: my OCD says let's change it, but it will require some backend changes https://github.com/EveripediaNetwork/ep-api/blob/main/src/Indexer/Validator/validator.service.spec.ts#L32 & https://github.com/EveripediaNetwork/ep-api/pull/132/files#diff-cb40f395074813c3980293999fb4f1c5c76d8d14873e1d7e3fe2db09a62e2b49R24

I would say lets revert it for now and we can do a multiple PR in the future

kesar (approved):

src/types/Wiki.ts
Royal-lobster: Done ! I reverted it back to underscores for now

PR #326, opened 2022-06-08
UI/UI Fixes for activity and wiki page

codiejay:
Update lineheight for wikiContent on wiki and wikiEdit. 
Also fixes wiki alignment bug plus padding and width. 

image
before

image
After

image
before: 

image
after 

image
before: 

image
after:

## Linked issues

closes #394, closes #394, closes https://github.com/EveripediaNetwork/issues/issues/394


kesar:
I think you linked wrong issue

codiejay:
> I think you linked wrong issue

Oopps; fixed this, thanks for the headsup.

Simplypheyie96 (changes requested):
image
the paddings on both sides of every wiki are yet to be added.

Also, check out this section of the wiki. there should be a little bit of space here before the footer starts

Simplypheyie96:
@kesar, concerning the issue of line height, are you seeing a difference on your screen. yours is quite larger and can be easily visible

Simplypheyie96:
This is the design on figma. notice the paddings on both sides. That is what every wiki should look like
image

src/pages/revision/[id].tsx
```diff
@@ -153,19 +190,53 @@ const Revision = () => {
...
+                        mobileView={false}
+                      />
                       <WikiInsights
                         dateTime={wiki.datetime}
                         wiki={wiki.content[0]}
                         ipfs={wiki.ipfs}
                       />
                     </Flex>
+                    <Flex
+                      display={{ lg: 'none', base: 'block', md: 'none' }}
```
Comment (anubra266) [inline src/pages/revision/[id].tsx:213]: If `md` and `lg` are the same, you don't need `lg`, because it's a higher breakpoint than `md`
```suggestion
                      display={{ base: 'block', md: 'none' }}
```

src/pages/revision/[id].tsx
```diff
@@ -153,19 +190,53 @@ const Revision = () => {
...
+                      flexDirection={{
+                        base: 'column-reverse',
+                        lg: 'row',
+                        md: 'row',
```
Comment (anubra266) [inline src/pages/revision/[id].tsx:217]: Same here, no need for `lg`
```suggestion
                        md: 'row',
```

src/pages/revision/[id].tsx
```diff
@@ -153,19 +190,53 @@ const Revision = () => {
...
+                      </Heading>
+                      <WikiInsights
+                        dateTime={wiki.datetime}
+                        wiki={wiki.content[0]}
+                        ipfs={wiki.ipfs}
+                      />
+                      <WikiMainContent wiki={wiki.content[0]} mobileView />
+                    </Flex>
+                    <Flex
+                      display={{ base: 'block', lg: 'none', md: 'none' }}
```
Comment (anubra266) [inline src/pages/revision/[id].tsx:231]:
```suggestion
                      display={{ base: 'block', md: 'none' }}
```

src/pages/wiki/[slug]/index.tsx
```diff
@@ -9,24 +9,53 @@ import {
...
 import { useAppSelector } from '@/store/hook'
 import { WikiHeader } from '@/components/SEO/Wiki'
 import { getWikiSummary } from '@/utils/getWikiSummary'
 import { getWikiImageUrl } from '@/utils/getWikiImageUrl'
+import RelatedMediaGrid from '@/components/Wiki/WikiPage/InsightComponents/RelatedMedia'
+import { RelatedWikis } from '@/components/Wiki/WikiPage/InsightComponents/RelatedWikis'
+import TwitterTimeline from '@/components/Wiki/WikiPage/InsightComponents/TwitterTimeline'
+
+const MobileMeta = (wiki: {
+  metadata: any[]
```
Comment (anubra266) [inline src/pages/wiki/[slug]/index.tsx:30]:
```suggestion
type Meta = {
  id: CommonMetaIds
  value: string
}

const MobileMeta = (wiki: {
  metadata: Meta[]
```

src/pages/wiki/[slug]/index.tsx
```diff
@@ -99,9 +128,46 @@ const Wiki = () => {
...
                   </Flex>
+                  {wiki.media && wiki.media.length > 0 && (
+                    <RelatedMediaGrid media={wiki.media} />
+                  )}
+                  <Flex
+                    display={{ base: 'block', lg: 'none', md: 'none' }}
+                    flexDirection="column"
+                  >
+                    <MobileMeta
+                      metadata={wiki.metadata}
```
Comment (anubra266) [inline src/pages/wiki/[slug]/index.tsx:165]: If the type defined gets disturbing, you can do
```suggestion
                      metadata={wiki.metadata as Meta[]}
```

anubra266 (changes requested):

src/pages/revision/[id].tsx
Comment (codiejay) [reply src/pages/revision/[id].tsx:213]: So the thing is for this UI we need to specify for `MD`. I will explain this offline

Simplypheyie96 (approved):
great work. @codiejay

kesar (approved):

PR #327, opened 2022-06-09
Track actions in ga

Emmanueldmlr:
# Track Actions in GA

- [ ] OPEN DRAWER
- [ ]  LOGIN_ATTEMPT ( w which provider )
- [ ]  LOGIN_SUCCESS
- [ ]  LOGIN_ERROR
- [ ]  SUBMIT_WIKI
- [ ]  SUBMIT_WIKI_ERROR ( w error )
- [ ]  Extra: new page on scroll infinite
- [ ]  Search items

## How should this be tested?

1. Try to perform the actions above.
2. those actions will be logged to the GA dashboard

fixes https://github.com/EveripediaNetwork/issues/issues/296

### DIFF
--- a/src/services/search/utils.ts
+++ b/src/services/search/utils.ts
@@ -82,7 +82,7 @@ export const useNavSearch = () => {
       logEvent({
         action: 'SEARCH_ITEM',
         params: {
-          data:query,
+          data: query,
         },
       })
       debouncedFetchResults(query, res => {
--- a/src/services/search/utils.ts
+++ b/src/services/search/utils.ts
@@ -7,6 +7,7 @@ import { getTagWikis } from '@/services/wikis'
 import { store } from '@/store/store'
 import { Category } from '@/types/CategoryDataTypes'
 import { Tag, WikiPreview } from '@/types/Wiki'
+import { logEvent } from '@/utils/googleAnalytics'
 import { debounce } from 'debounce'
 
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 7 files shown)


src/services/search/utils.ts
```diff
@@ -78,6 +79,12 @@ export const useNavSearch = () => {
   useEffect(() => {
     if (query && query.length >= 3) {
       setIsLoading(true)
+      logEvent({
```
kesar, line 82: we dont need search per key. we need to log final search. Its not meaningful to store that somebody searched for emm if they want to search for emmanuel.

src/utils/create-wiki.ts
```diff
@@ -189,46 +190,80 @@ export const useGetSignedHash = (deadline: number) => {
         } else {
           setIsLoading('error')
           setMsg(errorMessage)
+          logEvent({
```
kesar, line 193: api now returns errors so that's what we need to log: https://github.com/EveripediaNetwork/ep-api/pull/130#discussion_r889140609

kesar: that's also related w https://github.com/EveripediaNetwork/issues/issues/400 that should catch that error and show it in the UI

src/components/Profile/Collected.tsx
```diff
@@ -32,6 +33,7 @@ const Collected = () => {
             }),
           )
           if (result.data && result.data?.length > 0) {
+            pageView(`${router.asPath}/${fetchOffset}`)
```
kesar:
```suggestion
            pageView(`${router.asPath}/?page=${fetchOffset}`)
```

its probably better if we do it as a url param, that way in the future we might actually allow people to accessing those routes to visit that page

kesar (approved):

PR #328, opened 2022-06-09
fixed user profile problem

OleanjiKingCode:
# Fixed wikis size in user profile 

_This solves the arbitrary way the wiki sizes are when less than 3 so now they are like the design 

## How should this be tested?
By creating at least one wiki and checking it in the user profile

1. Before 
Screenshot (245)
Screenshot (239)

3. After

Screenshot (247)

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/405

### DIFF
--- a/src/components/Profile/Collected.tsx
+++ b/src/components/Profile/Collected.tsx
@@ -1,6 +1,5 @@
 import useInfiniteScroll from 'react-infinite-scroll-hook'
 import { FilterLayout } from '@/components/Profile/FilterLayout'
-import { useProfileContext } from '@/components/Profile/utils'
 import { getUserWikis } from '@/services/wikis'
 import { Center, SimpleGrid, Text, Spinner } from '@chakra-ui/react'
 import React, { useState, useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/Profile/Collected.tsx
```diff
@@ -72,9 +70,12 @@ const Collected = () => {
       )}
       <SimpleGrid
         ref={sentryRef}
-        minChildWidth={displaySize}
-        w="full"
-        spacing="4"
+        maxW="1050px"
+        w="100%"
+        mx="auto"
+        columns={[1, 2, 3]}
```
codiejay, line 76: We might want to keep this responsive object uniform across the codebase. 
Currently we might want to do it this way:  `{{base: 1, md: 2, xl: 3}}`. This was we keep it uniform and increase readability.

codiejay (changes requested):

src/components/Profile/Collected.tsx
kesar:
```suggestion
        columns={{base: 1, md: 2, lg: 3}}
```

@codiejay remember you can suggest if its a one line change so the PR creator can commit the change with one click 👍🏻

codiejay: Ah thanks @kesar  - will keep that in mind!!

PR #331, opened 2022-06-14
Refactor/infinite scroll

anubra266:
Resolves EveripediaNetwork/issues#420


anubra266:
Thanks @Royal-lobster It was because profile is supposed to be set to loading by default, different from others. Fixed it.

Royal-lobster (approved):

PR #332, opened 2022-06-14
Fixes the user profile

OleanjiKingCode:
# Fixed wikis size in user profile 

_This solves the arbitrary way the wiki sizes are when less than 3 so now they are like the design 

## How should this be tested?
By creating at least one wiki and checking it in the user profile

1. Before 
Screenshot (245)
Screenshot (239)

3. After

Screenshot (250)

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/405


kesar:
if the code review just have a minor code suggestion, I would recommend to approve + add the suggestion to avoid unblocking the merge

Simplypheyie96 (approved):

anubra266 (approved):

PR #339, opened 2022-06-17
Fixes Draft wikis causing runtime error

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/430


kesar (approved):

kesar:
good that we have sentry now :P

Royal-lobster:
> good that we have sentry now :P keep an eye there from time to time to catch possibles issues plz

Sure !! will monitor it for issues

PR #352, opened 2022-06-23
Stop overflow in create wiki highlight bar. 

codiejay:
The overflow was caused by us using a `vh` for the height of the bar: 

image
Before 

image
After

## Linked issues

closes #441, closes #441, closes https://github.com/EveripediaNetwork/issues/issues/441

### DIFF
--- a/src/components/Layout/Editor/Highlights/Highlights.tsx
+++ b/src/components/Layout/Editor/Highlights/Highlights.tsx
@@ -77,7 +77,7 @@ const Highlights = ({ initialImage, isToResetImage }: HightLightsType) => {
       borderColor="borderColor"
       borderRadius="7px"
       padding="15px"
-      h="75vh"
+      h="auto"
     >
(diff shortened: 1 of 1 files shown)


kesar:
is that generating a vertical scroll on the right side? the goal is to avoid that scroll. 

It should resize components inside of the right side (image most likely) to avoid that scroll. Make sure editors dont face that vertical scroll because it ruins the usability of creating wikis

codiejay:
> is that generating a vertical scroll on the right side? the goal is to avoid that scroll.
> 
> It should resize components inside of the right side (image most likely) to avoid that scroll. Make sure editors dont face that vertical scroll because it ruins the usability of creating wikis

No it is not. Every thing sits on that section and they don't need to scroll down to click edit.

kesar:
awesome! 👏🏻

kesar (approved):

PR #364, opened 2022-06-27
Switch network mobile

Emmanueldmlr:
# Rework Switch network modal

_Switch network modal should come up when the user is logged in and not connected to the right chain id.

## How should this be tested?

1. logout and switch your Metamask network to something else. The switch network modal shouldn't come up because you are not logged in.
2. log in to your account ensuring you are on the wrong network. the switch network modal should come up because you are logged in but on the wrong network. By clicking on the switch network, you will be prompted to switch your network

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/452


kesar (approved):

Softdev1 (approved):
we are good

PR #366, opened 2022-06-28
Syncs ```alpha``` branch with ```main```

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/455

### DIFF
--- a/.github/workflows/alpha-sync.yml
+++ b/.github/workflows/alpha-sync.yml
@@ -1,0 +1,35 @@
+name: Merge Main Into Alpha
+
+on:
+  push:
+    branches:
+      - 'main/*'
+
+jobs:
+  main:
+    name: Create PR Release to Alphas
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-alpha.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    name: Syncing branches
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Opening pull request
```
kesar, line 18: i think this will only open pull request, there is in the original specs an example of automerging as well

Royal-lobster: yeah i realized it late  😅 i am reverting that commit

PR #378, opened 2022-06-28
Strip html on Copy/Paste

codiejay:
1. Previously when you paste a hypertext it comes with the opening and closing elements which could be buggy especially with `anchor` tags. This PR address that by stripping away the tags. 
2. `<br/>`, `<br></br>` and `<br>` are now replaced with a line break when pasted into the Editor. 
3. Also focusing on search bar using the `/` has been disabled as it doesn't allow users to type `/` inside of the Editor. Now Search bar can be focused by clicking on `CMD + K` 

Before: 
image
 
After:  
image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/393
fixes https://github.com/EveripediaNetwork/issues/issues/463

### DIFF
--- a/src/components/Layout/Editor/Editor.tsx
+++ b/src/components/Layout/Editor/Editor.tsx
@@ -101,7 +101,10 @@ const Editor = ({ onChange, markdown = '' }: EditorType) => {
     const currentMd = editorRef.current
       ?.getInstance()
       .getMarkdown()
-      .toString() as string
+      .toString()
+      .replace(/<[^>]+>/gm, '')
+      .replace(/\\/g,'')
+
     if (markdown !== currentMd) {
       if (
         markdown.substring(0, EditorContentOverride.KEYWORD.length) ===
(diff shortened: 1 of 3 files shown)


src/components/Layout/Editor/Editor.tsx
```diff
@@ -101,7 +101,10 @@ const Editor = ({ onChange, markdown = '' }: EditorType) => {
     const currentMd = editorRef.current
       ?.getInstance()
       .getMarkdown()
-      .toString() as string
+      .toString()
+      .replace(/<[^>]+>/gm, '')
```
kesar, line 105: can you try to add a first replace of <br> or <br /> or <br/> for \n ? 

that way if someone copy paste something w line jumps we would keep the format. If its hard we can skip it, but it could be a nice feature

codiejay: So we shouldn't be replacing `line jumps <br/>` ?

kesar: we should be replacing them for \n

src/pages/create-wiki/index.tsx
```diff
@@ -336,17 +336,22 @@ const CreateWikiContent = () => {
     val: string | undefined,
     isInitSet?: boolean,
   ) => {
+    const newVal = val
```
Royal-lobster, line 339:
```suggestion
    const filteredContent = val
```

renaming so it makes it easy to read what it is doing

kesar (approved):

Royal-lobster (approved):

PR #388, opened 2022-06-29
uses user address if ens name does not exist

Emmanueldmlr:
# Use user wallet address if ens name does not exist

_switch from user unnamed to their address ( or ens address if the user has ). 99% of users will be unnamed so it's better to show there their address.
Before 
image

After
image

fixes https://github.com/EveripediaNetwork/issues/issues/469

### DIFF
--- a/src/components/Profile/LoadingProfile.tsx
+++ b/src/components/Profile/LoadingProfile.tsx
@@ -78,9 +78,6 @@ export const LoadingProfile = (props: LoadingProfileProps) => {
               <chakra.span fontWeight="medium">userName</chakra.span>
             </Skeleton>
           </Flex>
-          <Skeleton>
-            <chakra.span color="gray.500">Joined November 2020</chakra.span>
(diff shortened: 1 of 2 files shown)


kesar (approved):

kesar:
title: `${username || 'Unnamed'} Profile Page - Everipedia`,
          description: `${username || 'Unnamed'} profile page`,

search by Unnamed, I think we should replace it everywhere

Simplypheyie96 (commented):
It looks great. Well done.
little adjustment on mobile. The "joined in November" under the user's address is not properly centered.
image

kesar:
> It looks great. Well done. little adjustment on mobile. The "joined in November" under the user's address is not properly centered. image

we need to remove it since that data its not coming from anywhere, its demo data and we are not storing that on db

anubra266 (approved):

kesar (approved):

PR #395, opened 2022-07-01
Fix mobile issues

Emmanueldmlr:
#Fixes Mobile issues.

homepage:
- [x] trending wikis not visible
menu:
- [x] click in activity and menu does not close
activity page:
- [x] huge padding in header (reduce considerable)
- [x] huge text in header (check how big are headers in categories)
- [x] title of cards gets collapsed after 6 letters (it should show way more, it has more importance than the edited word for example)
## How should this be tested?
-Try to test each update as requested in the task
After
image
image

## Linked issues
fixes https://github.com/EveripediaNetwork/issues/issues/467


Simplypheyie96 (changes requested):
They look good. 
For the activity page, we might as well reduce font size of it's headings to like 14. so that more texts would be visible before collapse. @Emmanueldmlr

Softdev1 (approved):

PR #432, opened 2022-07-12
wiki markdown stripped

OleanjiKingCode:
# PR title goes here

_when a wiki is created and there's no summary written for it....in its card it shows a bit of markdown text which cannot be understood by users but now it has been removed 

## How should this be tested?

1. ## Before
  By creating a wiki without a summary and using some markdown features at the beginning of the wiki.. then after creation the wiki card shows a bit of markdown text 

2. ## Now 
 The markdown text has been removed from a wiki without summary.

## Linked issues
closes https://github.com/EveripediaNetwork/issues/issues/475


src/utils/getWikiSummary.ts
```diff
@@ -1,15 +1,16 @@
 import { Wiki, WikiPreview } from '@/types/Wiki'
 import { shortenText } from './shortenText'
+import RemoveMarkdown from 'remove-markdown'
 
 export enum WikiSummarySize {
   Small = 65,
   Medium = 70,
-  Big = 160,
+  Big = 250,
```
Royal-lobster, line 8: @OleanjiKingCode this should remain same as 160.

OleanjiKingCode: Done

src/utils/getWikiSummary.ts
```diff
@@ -1,15 +1,16 @@
...
 }
 
 export const getWikiSummary = (
   wiki: Partial<Wiki> | WikiPreview,
   size: WikiSummarySize = WikiSummarySize.Big,
 ) => {
-  return shortenText(wiki.summary || wiki.content || '', size)
+  return RemoveMarkdown(shortenText(wiki.summary || wiki.content || '', size))
```
Royal-lobster, line 15: instead you can do something like this
```javascript
if(wiki.summary)
      return shortenText(wiki.summary)
if(wiki.content){
const trimmedContent = shortenText(wiki.content, size + 100)
const cleanedContentSummary =  RemoveMarkdown(trimmedContent)
return cleanedContentSummary
}

```

Royal-lobster: by this you are not running the RemoveMarkdown function on entire long content but a shorter version but still accounting that there may be some markdown syntax trimmed at end which may not be cleaned properly so we are adding a 100 characters buffer and trimming back to normal size. once play with it to see if it works as expected 🤞

Royal-lobster (approved):

Royal-lobster (approved):

PR #468, opened 2022-07-20
Collapsing some wiki widget in the Mobile wiki page content

OleanjiKingCode:
# Collapsing some wiki widget in the Mobile wiki page content

_Since our priority should be the getting the user to see the wiki content quickiy i collapsed the widgets above the content

## How should this be tested?
By opening up a wiki on mobile

1. Before   
image

2. Now 
image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/511

### DIFF
--- a/src/components/Wiki/WikiAccordion/index.tsx
+++ b/src/components/Wiki/WikiAccordion/index.tsx
@@ -27,7 +27,7 @@ const WikiAccordion = ({
   ...rest
 }: BoxProps & AccordionProps) => {
   const { isOpen, onToggle } = useDisclosure()
-  const isDefaultOpen = useBreakpointValue({ base: base, xl: xl })
+  const isDefaultOpen = useBreakpointValue({ base, xl })
 
   return (
     <Box
--- a/src/components/Wiki/WikiAccordion/index.tsx
+++ b/src/components/Wiki/WikiAccordion/index.tsx
@@ -6,22 +6,26 @@ import {
   IconButton,
   useDisclosure,
   BoxProps,
+  useBreakpointValue,
 } from '@chakra-ui/react'
 import { RiArrowDownSLine } from 'react-icons/ri'
 
 interface AccordionProps {
   title: string
   withNoDarkBg?: boolean
+  IsOpen: boolean
   children: React.ReactNode
 }
 
 const WikiAccordion = ({
   title,
   withNoDarkBg,
+  IsOpen,
   children,
   ...rest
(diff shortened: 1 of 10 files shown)


kesar:
i would just collapse the commit message. main pic is powerful to get context of the wiki, i would not remove that as first element

OleanjiKingCode:
> i would just collapse the commit message. main pic is powerful to get context of the wiki, i would not remove that as first element





Sir what do you mean by collapsing?  i dont really get it

OleanjiKingCode:
> i would just collapse the commit message. main pic is powerful to get context of the wiki, i would not remove that as first element





Sir what do you mean by collapsing?  i dont really get it

kesar:
Captura de Pantalla 2022-07-20 a las 18 46 38

this by default only in mobile

kesar:
there are a bunch of widgets collapsed on desktop and mobile that were not in the specs.

The specs mentioned the commit msg. Some like twitter or related widgets sounds like a good idea to be collapsed by default on mobile, so feel free to keep it, but no reason for that on desktop since we have a bunch of space to fill on the screen 👍🏻

OleanjiKingCode:
> there are a bunch of widgets collapsed on desktop and mobile that were not in the specs.
> 
> The specs mentioned the commit msg. Some like twitter or related widgets sounds like a good idea to be collapsed by default on mobile, so feel free to keep it, but no reason for that on desktop since we have a bunch of space to fill on the screen 👍🏻

yes sir im sorry i missed that... done now tho

src/components/Wiki/WikiAccordion/index.tsx
```diff
@@ -6,22 +6,28 @@ import {
...
   BoxProps,
+  useBreakpointValue,
 } from '@chakra-ui/react'
 import { RiArrowDownSLine } from 'react-icons/ri'
 
 interface AccordionProps {
   title: string
   withNoDarkBg?: boolean
+  base: boolean
+  xl: boolean
```
kesar, line 17: this naming should change. Nobody will know without checking implementation what's the meaning of xl or base.

to follow chakra style of props we could send an object of breakpoints and define if we want them collapsed or not. make it optional and default to non-collapsed if prop is not defined.

```
      <WikiAccordion title="Media" collapsed={{base: false, xl: false}}> ...
```

kesar (changes requested):

kesar (approved):

PR #471, opened 2022-07-21
Dynamic color for icon on Profile Summary

codiejay:
image Before 

image After 

## Linked issues

closes #522, closes #522, closes https://github.com/EveripediaNetwork/issues/issues/522


Simplypheyie96 (commented):
it is kind of too bright compared to the rest of the elements. i used White alpha-700 in the design. Also, the Instagram and twitter icon seems a bit out of alignment with the others
image

Softdev1 (approved):

src/components/Layout/Editor/Highlights/HighlightsModal/HighlightsModal.tsx
```diff
@@ -50,7 +50,7 @@ export const LINK_OPTIONS = [
   {
     id: CommonMetaIds.EMAIL_URL,
     label: 'Email',
-    icon: <MdEmail />,
+    icon: <EmailIcon />,
```
Royal-lobster, line 53: @codiejay great work ! small suggestion, instead of fixing the colors styles directly inside icons, i believe it would be simpler if you create something like 

```javascript
const socialIconsStyles = {
    _dark: {{ fill: 'white', _hover: { fill: 'brand.500' } }}
    _light: {{ fill: '#4A5568', _hover: { fill: 'brand.500' } }}
}
```

and then pass the styles to all icons like this
```javascript
icon: <EmailIcon {...socialIconsStyles} />
```

that way we can simply change the common styles

anubra266: @Royal-lobster @codiejay The best way to go about it is do `<Icon fill="currentColor" />` in the icon itself. THis way, whatever you use as color becomes the `fill`. So afterwards you can use `_dark={{color:"..."}}` as usual, or just use semanticTokens. `color="colorToken"`

src/components/Icons/telegramIcon.tsx
```diff
@@ -3,12 +3,14 @@ import { Icon, IconProps } from '@chakra-ui/react'
 
 const TelegramIcon = (props: IconProps) => {
   return (
-    <Icon viewBox="0 0 20 20" fill="none" {...props}>
+    <Icon
+      viewBox="0 0 20 20"
+      _dark={{ fill: 'white', _hover: { fill: 'brand.500' } }}
+      _light={{ fill: '#4A5568', _hover: { fill: 'brand.500' } }}
```
anubra266, line 9: We should'nt do this in our icons. It makes them hardly reusable.

src/components/Layout/Editor/Highlights/HighlightsModal/HighlightsModal.tsx
Royal-lobster: @anubra266 this is cool ! so currentColor takes the context of color prop passed to the component right ?

anubra266: @Royal-lobster It's not really a react prop context thing, it's pure html svg css functionality. It takes whatever it finds as the color style value and uses it as `currentColor`

Royal-lobster: Ooooo i don't know that. thanks for the info 😇

Simplypheyie96 (approved):

anubra266 (approved):

anubra266 (changes requested):
CleanShot 2022-07-22 at 14 35 16@2x
The vertical spacing is bad, you could use a `Wrap` component instead of `HStack` and add a `spacingY` attribute

Royal-lobster:
@codiejay the icons need some gaps between rows

CleanShot 2022-07-22 at 19 02 49@2x

src/components/Wiki/WikiPage/InsightComponents/ProfileSummary.tsx
```diff
@@ -52,19 +53,26 @@ const SocialProfiles = ({
                   return item.id !== 'website' && item.id !== 'contract_url'
                 })
                 .map((social: { value: string; id: string }, i: any) => {
```
anubra266, line 55: Let's change the `any` type here. `i` is a number

Simplypheyie96:
> CleanShot 2022-07-22 at 14 35 16@2x
> 
> The vertical spacing is bad, you could use a `` component instead of `` and add a `spacingY` attribute

oh wow. i didn't see this because it's a line on my end.

anubra266 (changes requested):
CleanShot 2022-07-22 at 16 19 06@2x
This looks weird, let's use a facebook svg that starts the text at the beginning, this particular one is usually used with a background, but since we're not doing that, we shoud change it.

codiejay:
> CleanShot 2022-07-22 at 16 19 06@2x
> 
> This looks weird, let's use a facebook svg that starts the text at the beginning, this particular one is usually used with a background, but since we're not doing that, we shoud change it.

@Simplypheyie96 is it possible you get us an icon like so?

Simplypheyie96:
> > CleanShot 2022-07-22 at 16 19 06@2x
> > This looks weird, let's use a facebook svg that starts the text at the beginning, this particular one is usually used with a background, but since we're not doing that, we shoud change it.
> 
> @Simplypheyie96 is it possible you get us an icon like so?

i don't recall understand what you mean by a facebook icon that starts it's text at the beginning. you mean like this:
image

codiejay:
> > > CleanShot 2022-07-22 at 16 19 06@2x
> > > This looks weird, let's use a facebook svg that starts the text at the beginning, this particular one is usually used with a background, but since we're not doing that, we shoud change it.
> > 
> > 
> > @Simplypheyie96 is it possible you get us an icon like so?
> 
> i don't recall understand what you mean by a facebook icon that starts it's text at the beginning. you mean like this: image

image

This is the Facebook Icon I see on Figma. Could you send me this one you just posted?

Simplypheyie96:
it's on figma now. i just changed it

Simplypheyie96 (approved):
perfect!

anubra266 (approved):

PR #491, opened 2022-07-24
Fixing multi-select on Search Dropdown

codiejay:
The issue was the IDs of the items were identical and we were using this as the Key for the items. 
To fix this, we now have the the Key to be a combination of ID and the Unique UpdateTimes. 

The markdown symbols have also been replaced using Regex. 

image
Before.  

image
After. 

## Linked issues

closes #555, closes #555, closes https://github.com/EveripediaNetwork/issues/issues/555

### DIFF
--- a/src/components/Layout/Navbar/NavSearch/index.tsx
+++ b/src/components/Layout/Navbar/NavSearch/index.tsx
@@ -132,7 +132,7 @@ export const NavSearch = (props: NavSearchProps) => {
           article.images && article.images[0].id
         }`
         const value = fillType(article, SEARCH_TYPES.ARTICLE)
-        //This negates the bug that is casued by two wikis with the same title.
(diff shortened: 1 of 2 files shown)


kesar:
pls add screenshot of before and after.

why u didnt use the markdown strip code that we have in cards?

i think the wikis have different id, but same name, so make sure you can click in both and you go to different wiki pages

codiejay:
> pls add screenshot of before and after.
> 
> why u didnt use the markdown strip code that we have in cards?
> 
> i think the wikis have different id, but same name, so make sure you can click in both and you go to different wiki pages

I just realised that the issue is like unlike other parts where we fetch wikis, the wikis here had their Title as their ID. 
Please refer to image: 
image 
I'll find a way to fix this.

kesar (approved):

PR #498, opened 2022-07-25
Fixs magic connector chain id evaluating as string 

Royal-lobster:
# Changes
- fixes chain id and active id mismatch error due to chainId passed as string instead of number
- removes connector and connectors types for inferring types implicitly (which is possible by updating magic connector with its latest internal wagmi/core dependency)

# Screenshot
CleanShot 2022-07-26 at 00 36 54@2x


kesar:
👍 merging to avoid editors having issues

PR #521, opened 2022-08-01
Activity page images redesign

codiejay:
image Before

image After

 

## Linked issues

closes #00, closes #595, closes https://github.com/EveripediaNetwork/issues/issues/595

### DIFF
--- a/src/components/Activity/ActivityCard.tsx
+++ b/src/components/Activity/ActivityCard.tsx
@@ -1,5 +1,14 @@
 import React from 'react'
-import { HStack, Heading, Text, Box, Tag, Flex, Stack, AspectRatio } from '@chakra-ui/react'
+import {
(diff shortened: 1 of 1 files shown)


Simplypheyie96 (changes requested):
It looks good on desktop, but it should maintain the same spacing between card and text like it had before.
image

Also, the cards on mobile and medium break point shouldn't have the same sizes as the desktop.

Simplypheyie96 (changes requested):
font sizes of these guys aren't supposed to be the same as in md/mobile breakpoints as they are in desktop. it won't accommodate them
image
 
 The spacing between the title and subtitle is too much. the subtitle should be close to the title.
image

Simplypheyie96 (approved):

PR #581, opened 2022-08-24
Page View Counter

Royal-lobster:
Closes https://github.com/EveripediaNetwork/issues/issues/641

# Change
- added new mutation to rtk query for wiki page count
- added logic in wiki file to send the page count request to server
- added code to restrict page count increment by a timestamp at most 1 request for every hour for a wiki id

### DIFF
--- a/src/pages/wiki/[slug]/index.tsx
+++ b/src/pages/wiki/[slug]/index.tsx
@@ -4,7 +4,6 @@ import {
   getRunningOperationPromises,
   getWiki,
   getWikisByCategory,
-  usePostWikiViewCountMutation,
 } from '@/services/wikis'
 import { store } from '@/store/store'
 import { GetServerSideProps } from 'next'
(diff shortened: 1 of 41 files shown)


src/pages/wiki/[slug]/index.tsx
```diff
@@ -26,6 +27,24 @@ const Wiki = ({ wiki }: WikiProps) => {
 
   const toc = useAppSelector(state => state.toc)
 
+  const [incViewCount] = usePostWikiViewCountMutation()
+
+  useEffect(() => {
```
Comment (kesar) [inline src/pages/wiki/[slug]/index.tsx:32]: i know its still a draft but just in case, you could consider to extract later this into a custom hook to reduce LoC here, or at least to extract all logic to store /read pageviews to his own utils file 👍🏻

Comment (Royal-lobster) [reply src/pages/wiki/[slug]/index.tsx:32]: Ohh yes i am thinking about it. i will make a custom hook so it can also be used in revision page

kesar (approved):

PR #597, opened 2022-08-29
Fixes 900+ Redundant requests to ENS 

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/660

# Changes
- avatars in trending wiki cards are triggering ens lookup for each render. added a skip functionality for useENSData and using it in avatar component to avoid calling ens requests when it is not required.

# Links
- https://monopedia-ui-git-fix-900-requests-prediqt.vercel.app/

### DIFF
--- a/src/components/Elements/Avatar/Avatar.tsx
+++ b/src/components/Elements/Avatar/Avatar.tsx
@@ -30,7 +30,10 @@ const DisplayAvatar = ({
   size = 26,
   ...rest
 }: DisplayAvatarProps) => {
-  const [avatar, ,] = useENSData(address)
+  const [avatar, ,] = useENSData(
+    address,
+    avatarIPFS ? avatarIPFS?.length > 0 : false,
+  )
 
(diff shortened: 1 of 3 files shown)


Emmanueldmlr:
@Royal-lobster, can you explain the reason for the skip, maybe we could use it in IQ-UI also.

Royal-lobster:
> @Royal-lobster, can you explain the reason for the skip, maybe we could use it in IQ-UI also.

Since we cannot conditionally call hooks (in this case, we don't want to call it if avatarIPFS is present), i modified the hook to accept the information to not trigger the network request with in the hook itself.

Emmanueldmlr:
> > @Royal-lobster, can you explain the reason for the skip, maybe we could use it in IQ-UI also.
> 
> Since we cannot conditionally call hooks (in this case, we don't want to call it if avatarIPFS is present), I modified the hook to accept the information to not trigger the network request within the hook itself.

Oh, I understand now. Thanks

kesar (approved):

PR #620, opened 2022-09-02
Makes wiki page support ondemand static regeneration 

Royal-lobster:
# FOR WIKI/REVISION PAGES
[To be merged after api update]


kesar:
prob. we can get in this PR also revisions which are less likely to change.

also... let's make sure this does not cache dynamic stuff like price widgets or twitter feeds 👍🏻

PR #630, opened 2022-09-06
fix wagmi def

anubra266:
fix wagmi def


anubra266:
@kesar the wiki edits and settings should be working now

PR #685, opened 2022-09-12
Makes history page support ondemand static regeneration

Royal-lobster:
closes https://github.com/EveripediaNetwork/issues/issues/693
> requires change in backend before merge

# Changes
 - makes history page static


kesar:
nice! could you try to create the PR for backend as well? :P Ashakae splitted into his own service, so it might be not complex to do it :D

Royal-lobster:
> nice! could you try to create the PR for backend as well? :P Ashakae splitted into his own service, so it might be not complex to do it :D
> 
> pretty much few changes here: https://github.com/EveripediaNetwork/ep-api/blob/main/src/App/utils/revalidatePage/revalidatePage.service.ts

Cool 😄 i will try to create PR in api.

PR #700, opened 2022-09-14
Fixes alt text on cards

Adebesin-Cell:
Fixes alt text on cards

## How should this be tested?

1. Open [web](https://web.dev/measure/)
2. Include the vercel preview link
3. Check the accessibility report.

related to https://github.com/EveripediaNetwork/issues/issues/708

### DIFF
--- a/src/components/Elements/Avatar/Avatar.tsx
+++ b/src/components/Elements/Avatar/Avatar.tsx
@@ -63,7 +63,15 @@ const DisplayAvatar = ({
       />
     )
   } else if (avatar) {
-    content = <Avatar h={`${size}px`} w={`${size}px`} src={avatar} {...rest} />
+    content = (
+      <Avatar
+        h={`${size}px`}
+        w={`${size}px`}
+        src={avatar}
+        name={alt}
+        {...rest}
+      />
+    )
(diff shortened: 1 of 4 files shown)


kesar:
this looks good, but make sure all issues are solved. You could put them all in the same PR since they are pretty small, and we could review all at once, instead of creating 4 different pr and 4 different revisions 👍🏻 

it usually depends in how big are the issues. IF they are big you should split in different PR, if they are 5 - 10 lines changes you can save time bundle them into same PR

src/components/Elements/Avatar/Avatar.tsx
```diff
@@ -63,13 +63,21 @@ const DisplayAvatar = ({
       />
     )
   } else if (avatar) {
-    content = <Avatar h={`${size}px`} w={`${size}px`} src={avatar} {...rest} />
+    content = (
+      <Avatar
+        h={`${size}px`}
+        w={`${size}px`}
```
anubra266, line 69: When `width` and `height` is the same, you can use boxSize
```suggestion
        boxSize={`${size}px`}
```

anubra266 (approved):

kesar (approved):

PR #707, opened 2022-09-15
Perf improvements test PR

anubra266:
Experimental PR for approach on the state glitch issue for @Royal-lobster to co-brainstorm.


src/components/Wagmi/DynamicWagmiProvider.tsx
```diff
@@ -32,12 +21,11 @@ export const DynamicWagmiProvider = ({
...
   )
-  const Wrapper = isWagmiWrapped ? WagmiProvider : React.Fragment
   return (
     <>
-      <WagmiStatusContext.Provider value={value}>
-        <Wrapper>{children}</Wrapper>
-      </WagmiStatusContext.Provider>
+      <WagmiStatusProvider value={value}>
+        <WagmiProvider>{children}</WagmiProvider>
+      </WagmiStatusProvider>
```
Royal-lobster, line 28: @anubra266 i think here we are just loading wagmi all the time dynamically. but this is making the magic requests to disappear from the metrics not sure why

anubra266: @Royal-lobster, what do you mean by dissapearing from metrics?

Royal-lobster: also now the wagmi status context is not really doing much i think since we are loading the wagmi all times

Royal-lobster: > @Royal-lobster, what do you mean by dissapearing from metrics?

Yes ! it is no longer in metrics, before i did the dynamic import directly on _app.ts but it showed on metrics, but now it is now showing even tho we are doing same in different file, not sure why exactly

PR #739, opened 2022-09-23
Fixes italic markdown

Adebesin-Cell:
Fixes italic markdown

fixes https://github.com/EveripediaNetwork/issues/issues/751
fixes https://github.com/EveripediaNetwork/issues/issues/737

### DIFF
--- a/src/components/Wiki/WikiPage/WikiMainContent.tsx
+++ b/src/components/Wiki/WikiPage/WikiMainContent.tsx
@@ -3,6 +3,7 @@ import { Box, Heading, useColorMode } from '@chakra-ui/react'
 import React, { useMemo } from 'react'
 import ReactMarkdown from 'react-markdown'
 import remarkGfm from 'remark-gfm'
+import rehypeRaw from 'rehype-raw'
 import { store } from '@/store/store'
 import { addToTOC } from '@/components/Wiki/WikiPage/CustomRenderers/customHeadingRender'
 import { getWikiMetadataById } from '@/utils/getWikiFields'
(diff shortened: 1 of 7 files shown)


kesar (approved):

src/components/Wiki/WikiPage/WikiMainContent.tsx
```diff
@@ -33,6 +34,7 @@ const MarkdownRender = React.memo(({ wiki }: { wiki: Wiki }) => {
   return (
     <ReactMarkdown
       remarkPlugins={[remarkGfm]}
+      rehypePlugins={[rehypeRaw]}
```
kesar, line 37: ideally we should parse HTML only in editor. The problem its that this could lead into some html injection which could steal user credentials

kesar (changes requested):

Softdev1 (approved):

src/components/Wiki/WikiPage/WikiMainContent.tsx
Adebesin-Cell: @Royal-lobster and I are experimenting to work it out from the editor. Though I got a solution: 
Screenshot 2022-09-23 105832

kesar: ohh nice idea. then i think we could use it w that 👍🏻

Royal-lobster: @kesar we found the issue, it is that when user selects the text which has some charecters immediately to next to it, the editor seems to decorate it with html tags instead of traditional markdown syntax. we are looking for a fix for this so we no longer need to use hehype-raw. if we fail to find any we can proceed with rehype-raw + rehype-sanitize it also filters the iframes and other dangerous attributes on html tags 
 
image

Softdev1 (approved):

Softdev1 (approved):

Softdev1 (approved):

kesar (approved):

PR #744, opened 2022-09-24
Updating Static about page img. 

codiejay:
## Linked issues 

cc: @Softdev1 

closes #00, closes #001, closes https://github.com/EveripediaNetwork/issues/issues/731


Simplypheyie96 (changes requested):
image looks fine now. But the arrow are not yet right.
for the first section, the arrows should be placed in the middle not at the top.
And the arrows on the second section need to come down a little for it to look balanced. It should be placed in the middle in reference of the two circles on the sides.
same goes for other breakpoints.
image

Simplypheyie96 (approved):

PR #770, opened 2022-10-04
779 increase content description text on search on desktop

codiejay:
closes https://github.com/EveripediaNetwork/issues/issues/799


Royal-lobster (changes requested):
@codiejay make summary to give longer content too. it will make the empty space filled.

CleanShot 2022-10-04 at 20 20 13@2x

kesar:
plz if PR has been accepted try to merge it first thing in the morning unless you are planning to work on it @codiejay 

If we have multiple PR opened we can end up with conflicts, merging them fast avoid that 👍🏻

PR #775, opened 2022-10-04
Implement user can publish

Emmanueldmlr:
we might disable publish button and explain that it's not in the whitelist.

this will change in the future and be replaced w checking if the user has nft passport, so make sure you do code modular to be replaced later.

https://polygonscan.com/address/0xbe5d150acb420c7ea010c1f525733eb3c1d26fa1#readContract

you need to check the function isEditorWhitelisted(address)

fixes https://github.com/EveripediaNetwork/issues/issues/782


Softdev1:
The UI implementation seems really cool, in this way editors will not even get to get errors at all .. well-done @Emmanueldmlr

Simplypheyie96:
@Emmanueldmlr The tooltip should be in white and it's corresponding color in dark mode just like other tooltips on the site
image

PR #783, opened 2022-10-05
Avoids getWiki request for new wiki

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/788


kesar (approved):

kesar:
is there anything else to do in this PR? i think it looks good to avoid querying api. No rush but I was thinking if we can get it merge it before today's call :D

kesar (approved):

PR #828, opened 2022-10-10
Edited wiki empty page

Emmanueldmlr:
I realized that we don't have an empty page for the 'edits' tab on the user profile and it's just the same as the one for wikis.
But that might be confusing.
So an empty page that describes what is supposed to be there would be appropriate 

fixes https://github.com/EveripediaNetwork/issues/issues/806


Emmanueldmlr:
@Simplypheyie96, done now.

Simplypheyie96 (approved):
Can't view because i have edits, but i'm sure you did a great job

Emmanueldmlr:
Nice. I will just wait for @kesar approval before merge.

kesar (approved):

PR #831, opened 2022-10-11
Feat/minimal

JustinCast:
- [x] Update the image in the about section
- [x] Include another address for the Korean blog entries


kesar:
fixed in other prs

PR #844, opened 2022-10-11
Fixes about page revamp

Adebesin-Cell:
Fixes about page revamp

fixes https://github.com/EveripediaNetwork/issues/issues/735


Simplypheyie96 (changes requested):
break this part a bit more. it is too wide
image

kesar (approved):

Royal-lobster (approved):

PR #852, opened 2022-10-11
logs platform language change event

Emmanueldmlr:
logs platform language change event on GA

fixes https://github.com/EveripediaNetwork/issues/issues/814


kesar:
👍🏻

PR #873, opened 2022-10-17
Fixes ENS suggestion on settings input username.

Adebesin-Cell:
Fixes ENS suggestion on settings input username.

fixes https://github.com/EveripediaNetwork/issues/issues/810

### DIFF
--- a/src/components/Settings/ProfileSettings.tsx
+++ b/src/components/Settings/ProfileSettings.tsx
@@ -1,6 +1,5 @@
 import React, { FormEvent, useCallback } from 'react'
 import {
-  Text,
   Flex,
   Input,
   Textarea,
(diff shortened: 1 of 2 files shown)


src/components/Settings/ProfileSettings.tsx
```diff
@@ -206,7 +206,23 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
         <VStack flex="2" align="left" spacing={4}>
           {/* PROFILE: USER NAME */}
           <FormControl isRequired isInvalid={inputUsername.error !== ''}>
-            <FormLabel htmlFor="username">Username</FormLabel>
+            <Flex>
+              <FormLabel htmlFor="username">Username</FormLabel>
+              {userENSAddr && (
+                <Text
```
Royal-lobster, line 212: @Adebesin-Cell Make it a button with variant `unstyled` or `link` so we will get keyboard accessibility and pointer on hover.

src/components/Settings/ProfileSettings.tsx
```diff
@@ -206,7 +206,23 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
...
+                  fontSize="xs"
+                  mb="2"
+                  onClick={() => {
+                    setInputUsername({
+                      value: userENSAddr,
+                      error: validateUsername(userENSAddr),
+                    })
+                  }}
+                >
+                  (Click to use your ENS {userENSAddr})
```
Royal-lobster, line 222: i think we can remove the message or show some other message if the username is already the user's ENS.

CleanShot 2022-10-17 at 19 32 39@2x

Adebesin-Cell: I think removing the message is fine.

Softdev1: I have the same comment as that of @Royal-lobster  once the changes is made, we are good to go

Royal-lobster (approved):

PR #876, opened 2022-10-17
Fixes chakra external links

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/825

### DIFF
--- a/src/components/Profile/UserSocialLinks.tsx
+++ b/src/components/Profile/UserSocialLinks.tsx
@@ -23,7 +23,7 @@ const UserSocialLinks = ({ links, address }: ProfileLinksProps) => {
           return (
             <Link
               position="relative"
-              rel="nofollow"
+              isExternal
               target="_blank"
               href={dt[social[0]].urlPrefix(social[1] || '')}
             >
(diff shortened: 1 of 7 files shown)


src/components/Profile/UserSocialLinks.tsx
```diff
@@ -23,7 +23,7 @@ const UserSocialLinks = ({ links, address }: ProfileLinksProps) => {
           return (
             <Link
               position="relative"
-              rel="nofollow"
+              isExternal
```
kesar, line 26: i think isExternal only adds the noreferrer but we need the nofollow

Royal-lobster: Oh let me investigate this further, i guess chakra ui is not giving us the option to add nofollow, i think we can create our own chakra.a link possibly without using the import of `@chakra/system`

Royal-lobster: @kesar found it, its link overlay the issue - https://github.com/chakra-ui/chakra-ui/blob/95aa4499c72de28e91dd9274177c529db55e7964/packages/components/layout/src/link-box.tsx#L16 here the rest of the props are spread before the rel so it is getting overriden

kesar: makes sense 👍🏻

kesar (approved):

PR #899, opened 2022-10-22
Fix issue with userid in ga

Emmanueldmlr:
# Fix userId not showing in GA

fixes https://github.com/EveripediaNetwork/issues/issues/836

### DIFF
--- a/src/components/Layout/GoogleAnalyticsScripts.tsx
+++ b/src/components/Layout/GoogleAnalyticsScripts.tsx
@@ -20,8 +20,10 @@ const GoogleAnalyticsScripts = () => {
               send_page_view: false
             });
             ${
-              typeof getUserAddressFromCache() === 'string' && 
-              `gtag('set', 'user_id', ${JSON.stringify(getUserAddressFromCache())})`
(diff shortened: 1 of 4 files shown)


Emmanueldmlr:
@kesar, can you review this if I have done the right thing?

kesar:
yes, but make sure you put the branch in ready for review if you want a review bcs that way its easy to request changes 👍🏻

src/components/Layout/GoogleAnalyticsScripts.tsx
```diff
@@ -18,6 +19,11 @@ const GoogleAnalyticsScripts = () => {
               page_path: window.location.pathname,
               send_page_view: false
             });
+            gtag('set', 'user_id', ${
+              typeof getUserAddressFromCache() === 'string'
+                ? JSON.stringify(getUserAddressFromCache())
+                : 'anonymous'
```
kesar, line 25: nope, this would put all the non-logged user under the same account and that's not what we want. the set user_id should only happens if user has id

kesar:
👍🏻

PR #919, opened 2022-10-27
Category query

OleanjiKingCode:
# Randomizing categories shown in homepage

Using the array filter method and math . random I removed one random category from the category list on reload 

image

image

image


src/components/Landing/CategoriesList.tsx
```diff
@@ -18,6 +18,8 @@ interface CategoriesListProps {
 }
 const CategoriesList = ({ categories }: CategoriesListProps) => {
   const { t } = useTranslation()
+  const RandcategoryItem: string =
```
kesar, line 21: variable names are in camelCase

I think its gonna be more readable if you just:

const filteredCategories =  categories.splice(Math.floor(Math.random()*categories.length), 1);

and use that w current map

kesar: you haven solve this or reply the reason

OleanjiKingCode:
I didn't use the splice method due to an error I had while using It, I decided to just go with the one I initially used

kesar:
> I didn't use the splice method due to an error I had while using It, I decided to just go with the one I initially used

that's not a proper response to a code review. Please explain what error or why do you think your way its cleaner / better to the suggested one.

OleanjiKingCode:
@kesar ?

kesar (approved):
looking good

kesar:
wait

https://github.com/EveripediaNetwork/ep-ui/blob/main/src/data/AllCategoriesData.ts#L73

u are renaming a file

OleanjiKingCode:
> wait
> 
> https://github.com/EveripediaNetwork/ep-ui/blob/main/src/data/AllCategoriesData.ts#L73
> 
> u are renaming a file
Yes, I changed that because the link used to display the dapps image in prod doesn't work in dev so I changed the it now it should show in both dev and prod

kesar:
but then you need to rename it in the file i sent you, otherwise we will have a 404 image ther

kesar:
good 👍🏻

PR #921, opened 2022-10-28
Fix add tag problem

Emmanueldmlr:
when you type "en" and click on a result, it doesn't add the tag properly. If you search by "En" and click on the result it seems to work, so perhaps it's related w uppercases.

Typing "e" and clicking in developers don't seem to work, so something might be wrong there

fixes https://github.com/EveripediaNetwork/issues/issues/849


Emmanueldmlr:
https://user-images.githubusercontent.com/70170061/198618687-1158fbf4-9ad6-4ac6-bf60-4c1333ffed10.mov

PR #926, opened 2022-10-31
Ranking page

codiejay:
- [x] Create Ranking page 
- [x] Create UI in accordance to Figma design 
- [x] Fetch real data from endpoint. 

closes https://github.com/EveripediaNetwork/issues/issues/818


Simplypheyie96 (changes requested):
i guess you are still working on it but it is not done yet. the spacing between header and content is too much and there are still a lot of details that have been swapped or not properly executed yet. 
It is best to inspect the design on figma to make sure details are correct.
Also, the dark mode too is not properly done yet as well as other breakpoints.

codiejay:
> i guess you are still working on it but it is not done yet. the spacing between header and content is too much and there are still a lot of details that have been swapped or not properly executed yet. It is best to inspect the design on figma to make sure details are correct. Also, the dark mode too is not properly done yet as well as other breakpoints.

Just updated now. Thnks!

Simplypheyie96 (changes requested):
- [ ] Add background color for the header in light mode
image
- [ ] Align this properly like on figma and rework the font size/type and color for this part, both in light/dark mode. Inspect this part in figma again properly to see what I mean.
image
- [ ] reduce the distance between these guys too
image

Simplypheyie96 (changes requested):
Header looks like this on my 13inches laptop.
image
i think we have to do something about the numbers here to prevent it from breaking like this in mobile bp/tablet too
image
there are still some corrections to be made on the mobile bp/tablet
image

Simplypheyie96 (changes requested):
Tablet doesn't look good yet. 
image

Simplypheyie96 (changes requested):
This is all that is left i think. It is joined together here.
image

Simplypheyie96:
I think the ui looks good now

Royal-lobster:
CleanShot 2022-12-14 at 14 02 22@2x

i think the tables are not symmetric, probably because i am viewing from 13 inch. how about we center the tables so it looks good on all viewports ?

Softdev1:
I thought I pointed this out to you @codiejay  about the centralizing the table since they are just two and you said you would discuss with @Simplypheyie96  please ensure to discuss please

Simplypheyie96 (changes requested):
@codiejay change color in light mode to brand 500 for the prev/next buttons
image

then this page on md bp needs to fixed. 
image

Simplypheyie96 (changes requested):
Now the dark mode needs to be brand 800 for the buttons :(
image

src/components/Rank/RankCardItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+  const getTokenAlias = (text: string) => {
+    if (text === null) {
+      return 'Unknown'
+    }
+    if (text) {
+      return text
+    }
+    return ''
+  }
```
Royal-lobster, line 34: @codiejay since these both functions are same you can remove `getTokenAlias`. also it can be shortened to this 

```typescript
  const getAlias = (text: string) => (text ?? 'Unknown') || ""
```

src/components/Rank/RankCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+          </Button>
+          <Button
+            rightIcon={<AiOutlineDoubleRight />}
+            _hover={{ bg: 'transparent' }}
+            p="0"
+            bg="transparent"
+            _active={{ bg: 'transparent' }}
+            onClick={() => {
+              offsetIncrease()
+            }}
```
Royal-lobster, line 107:
```suggestion
            onClick={offsetIncrease}
```

just incase you don't know but you can give the function name to onClick, it calls it with minimum syntax !

src/components/Rank/RankCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+        useGetNFTRankingQuery({
+          kind: queryKind,
+          limit: 10,
+          offset: queryLimit,
+        }) // eslint-disable-next-line
+      : useGetTokenRankingQuery({
+          kind: queryKind,
+          limit: 10,
+          offset: queryLimit,
+        })
```
Royal-lobster, line 41: wow thats cleaver but i don't think we can use hooks that way 😅 i think its better if we create a custom hook in `services/ranking/utils.tsx` which takes in kind and gives the data. like this

```javascript
const useRankingQuery = ({kind, limit, offset}) => {
	const [data, setData] = useState()
	const [loading, setLoading] = useState(false)
	useEffect(()=>{
		setLoading(true);
		let data = []
		if(kind === 'NFT')
			data = (await store.dispatch(getNftRanking.initiate({kind, limit, offset}))).data
		else
			data = (await store.dispatch(getTokenRanking.initiate({kind, limit, offset}))).data
		setLoading(false);
	}, [])
	return {data, loading}
}
```

then later use the query in this place 

```javascript
const {data: nftsQuery, loading} = useRankingQuery({
	kind: queryKind,
	limit: 10,
	offset: queryLimit,
})
```

this should remove linting issues, flexible for new changes changes and also abstracts the logic well

src/components/Rank/InvalidRankCardItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+            >
+              $0.00
+            </Text>
+            <Flex
+              alignItems="center"
+              gap={0.5}
+              width="100%"
+              justifyContent="end"
+            >
+              {/* <AiFillCaretUp color={upIndicationIconColor} /> */}
```
Royal-lobster, line 66:
```suggestion
```

src/components/Rank/RankCard.tsx
codiejay: This is a way logical way to do it, but the issue we have is accessing the object inside of data. 
The dataset returned by NFTQuery is different from the query returned by Tokenquery. 

But my logic is a bit off especially with the lint issue. I will find a way incorporating your idea and fix this. 
Thanks again!

Royal-lobster: Hmm, i don't think it works since hooks should be called top level of a react component or another hook. is it working ?

PR #938, opened 2022-11-04
Rework tag implementation

Emmanueldmlr:
when you type "en" and click on a result, it doesn't add the tag properly. If you search by "En" and click on the result it seems to work, so perhaps it's related w uppercases.

Typing "e" and clicking in developers don't seem to work, so something might be wrong there

fixes https://github.com/EveripediaNetwork/issues/issues/849

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "axios": "^0.26.1",
     "bigpicture": "^2.6.1",
     "boring-avatars": "^1.6.3",
+    "chakra-react-select": "^4.4.0",
     "chakra-ui-steps": "^1.7.2",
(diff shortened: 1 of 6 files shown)


Emmanueldmlr:
@Simplypheyie96 and @Softdev1 , I reworked the tags selection component. Kindly help me test to be sure it works as expected and we have the same user experience.

To test: try to create a wiki and add tags to the wiki.

I will be waiting for your feedback. Thanks.

Simplypheyie96 (approved):
It works perfectly

kesar (approved):

PR #942, opened 2022-11-08
Iterations on Branding Page

codiejay:
- [x] Add "Branding" item to footer links. 
- [x] Add new logos and update sections. 

closes https://github.com/EveripediaNetwork/issues/issues/871


Simplypheyie96 (changes requested):
new sections looks good
one thing i wanted to point out that i totally forgot about is the quotes section. The texts have a bigger font size in the design. so it shouldn't be the same as the others.
image
Then i also thing that clicking outside the boxes should revert the download options to it's original state
image

spacing between the words are a bit much on mobile and tab
image

kesar:
> Then i also thing that clicking outside the boxes should revert the download options to it's original state

I would also suggest make them look like buttons if they are buttons. We need to train users to understand what buttons are in our site and use them across every single clickable interaction that will produce an effect.

current design looks like a text, not even a link. (a link should be pink w underline effect on hover, and it would be used if you are navigating to a page / url, in this case you are downloading a file)


corners in main image are not rounded like everywhere else. We should keep coherence and avoid squared corners if we use rounded.

image

codiejay:
Fixing these issues now. 
@Simplypheyie96 will you go on to update the design with the new linking UI @kesar is suggesting?

Simplypheyie96:
corners are rounded in the design . for the download button on the page, i felt it would be too colorful having everything in pink since the page itself is so colorful. so i added the text as black with a logo to indicate the download but i'll update that now

kesar:
> corners are rounded in the design . for the download button on the page, i felt it would be too colorful having everything in pink since the page itself is so colorful. so i added the text as black with a logo to indicate the download but i'll update that now
> 
> 

oh i see. we could overlay everything else grey if thats the issue, or as middle point we could try a design w white buttons that i think we have some in other pages. 

up to you, its not a big issue 👍👍

Simplypheyie96:
> > corners are rounded in the design . for the download button on the page, i felt it would be too colorful having everything in pink since the page itself is so colorful. so i added the text as black with a logo to indicate the download but i'll update that now
> > 
> > 
> 
> oh i see. we could overlay everything else grey if thats the issue, or as middle point we could try a design w white buttons that i think we have some in other pages. 
> 
> up to you, its not a big issue 👍👍

an idea that came to me now is to just remove the download text or put them inside the boxes themselves such that a click in the box will show the options to download in svg or png.

Simplypheyie96:
i used pink icons for the white backgrounds and white icons for the rest of the page due to contrast issues. I think it looks neater now compared to before. @kesar @codiejay 
image

Simplypheyie96 (changes requested):
-change color of download icons to brand color (pink)/ take note of color find dark mode
- The text here on the quotes should have a bigger font size compared to the other texts.
image
then increase spacings between hero and the quotes, then the quotes section and the first sub heading
image
- these texts should also be in pink. check figma
image

Simplypheyie96:
you missed out the last sentence and the 3rd sentence is duplicated @codiejay 
image

Simplypheyie96 (changes requested):
This should be gray700 in dark mode
image
image
there should be more spacing here
image

Simplypheyie96:
@codiejay additional iterations according to the meeting yesterday is to switch the IQ logo only to the top and bring down the logo with texts. That is, switch sections. 
like this:
image

Simplypheyie96 (approved):
well done. I think we can merge now

Simplypheyie96:
You can totally merge now @codiejay

src/pages/static/branding.tsx
```diff
@@ -92,8 +81,46 @@ const BrandingPage = () => {
...
+      bg: '/branding/braindaoaltdeepblack.png',
+      download: '/branding/downloadassets/logobraindoadeepblack',
+    },
+  ]
+  const heroImg = useColorModeValue('/brandingBrain.png', '/brandingBrainB.png')
   return (
-    <Box bg="tetiaryGray" _dark={{ backgroundColor: 'tetiaryDark' }} pb={20}>
+    <Box
+      m="0"
+      bg="tetiaryGray"
```
Adebesin-Cell, line 119: Use semantic tokens.

src/components/Branding/BrandingAssets.tsx
```diff
@@ -48,24 +64,44 @@ export const BrandingAssets = ({
...
+              setShowDownloadOptions(true)
+            }}
+            fontSize="0px"
+            bg="white"
+            boxShadow="0px 0px 4px 2px rgba(183, 183, 183, 0.21)"
+            variant="outline"
+            border="none"
+            display="flex"
+            borderRadius="200px"
+            m={{ base: '0', lg: '2', md: '2' }}
```
Adebesin-Cell, line 92:
```suggestion
            m={{ base: '0', md: '2' }}
```

src/components/Branding/BrandingAssets.tsx
```diff
@@ -48,24 +64,44 @@ export const BrandingAssets = ({
...
+            p="2"
+            w="fit-content"
+            alignItems="center"
+            justifyItems="center"
+            justifyContent="center"
+            rightIcon={
+              <BiCloudDownload fontSize="20px" color={downloadIconColor} />
+            }
+            iconSpacing="0"
+            transform={{ base: 'scale(0.5)', lg: 'scale(1)', md: 'scale(1)' }}
```
Adebesin-Cell, line 102:
```suggestion
            transform={{ base: 'scale(0.5)', md: 'scale(1)' }}
```

src/components/Branding/BrandingAssetDownloadBttn.tsx
```diff
@@ -25,7 +25,13 @@ const BrandingAssetDownloadBttn = ({
         width="fit-content"
         onClick={closeDownloadOptions}
       >
-        <Text textAlign="center">{text}</Text>
+        <Text
+          textAlign="center"
+          color="brand.500"
```
Adebesin-Cell, line 30: Use semantic tokens.

Adebesin-Cell (changes requested):

Adebesin-Cell (approved):
Nice @codiejay.

PR #965, opened 2022-11-21
iterations for branding page

codiejay:
UI iterations for branding page

closes https://github.com/EveripediaNetwork/issues/issues/886


Simplypheyie96 (changes requested):
the hero section is not yet there. the image is too close to the hero text. and the image feels too small
spacing here is still huge
image

Simplypheyie96 (approved):

Softdev1 (approved):
this is good now @codiejay

PR #968, opened 2022-11-22
Glossary iteration

Emmanueldmlr:
# Glossary iteration
- [ ]  Reduce the search bar placeholder text size and color
- [ ]  Reduce the size of the Tags
- [ ]  Reduce margin around Tags
- [ ] Autohide tags when scrolling
- [ ] Make the filter section collapsible when scrolling
- [ ] Reduce the glossary search result card header text

fixes https://github.com/EveripediaNetwork/issues/issues/883

### DIFF
--- a/src/components/Glossary/GlossaryWikiCard.tsx
+++ b/src/components/Glossary/GlossaryWikiCard.tsx
@@ -40,7 +40,7 @@ const GlossaryWikiCard = ({
         <VStack pb="2" w="full" px="3">
           <LinkOverlay href={`/wiki/${wikiId}`} w="full">
             <Heading
-              size={{ base: 'sm', lg: 'lg' }}
+              size={{ base: 'sm', lg: 'md' }}
               my="10px"
               color="brand.500"
             >
--- a/src/pages/glossary/index.tsx
+++ b/src/pages/glossary/index.tsx
@@ -29,7 +29,6 @@ import {
   // useGetTagsQuery,
 } from '@/services/glossary'
 import { useInView } from 'react-intersection-observer'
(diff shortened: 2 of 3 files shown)


Simplypheyie96 (changes requested):
i think we have to change this annoying feature of searching and the page scrolling down to the letter. That is kind of annoying if a user types one letter and the search bar disappears. not a good experience. I think it should bring suggestions up and not scroll down to where the frame with such letter is. check how search works on bit. 
Asides that, i think it looks good

Emmanueldmlr:
> i think we have to change this annoying feature of searching and the page scrolling down to the letter. That is kind of annoying if a user is types one letter and the search bar disappears. not a good experience. I think it should bring suggestions up and not scroll down to where the frame with such letter is. check how search works on bit. Asides that, i think it looks good

I guess we can fix the scroll issue in another PR, since we might have to brainstorm on how it should work.

OleanjiKingCode:
> i think we have to change this annoying feature of searching and the page scrolling down to the letter. That is kind of annoying if a user types one letter and the search bar disappears. not a good experience. I think it should bring suggestions up and not scroll down to where the frame with such letter is. check how search works on finance glossary page. Asides that, i think it looks good

Okay, what about, after bringing up suggestions like the normal search bar does  in IQ wiki then one is clicked it scrolls to the wiki on the page or should it go to the wiki page directly

OleanjiKingCode (approved):

Softdev1 (approved):

Softdev1:
we can brainstorm on the scroll issue and look at the possible solution.For now, I think we can merge this PR @Emmanueldmlr

Emmanueldmlr:
I'm merging this now.

PR #999, opened 2022-12-05
Refactoring code for admin

codiejay:
closes https://github.com/EveripediaNetwork/issues/issues/914

### DIFF
--- a/src/components/Admin/WikiCreatedInsight/InsightTableCreatedWiki.tsx
+++ b/src/components/Admin/WikiCreatedInsight/InsightTableCreatedWiki.tsx
@@ -38,7 +38,7 @@ type InsightTableWikiCreatedProps = {
 
 const TableHead = ({ text }: { text: string }) => {
   return (
-    <Th color="#718096" textTransform="none" fontWeight="medium">
+    <Th color="primaryGray" textTransform="none" fontWeight="medium">
       <Text fontWeight="bold">{text}</Text>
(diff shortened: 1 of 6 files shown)


src/components/Admin/WikiCreatedInsight/InsightTableCreatedWiki.tsx
```diff
@@ -36,6 +36,14 @@ type InsightTableWikiCreatedProps = {
   hideWikisFunc: () => void
 }
 
+const TableHead = ({ text }: { text: string }) => {
+  return (
+    <Th color="#718096" textTransform="none" fontWeight="medium">
```
Emmanueldmlr, line 41: I think it will be better to create a semantic token for the color.

src/components/Admin/WikiEditorInsight/InsightTableWikiEditors.tsx
```diff
@@ -65,6 +65,14 @@ type InsightTableWikiEditorsProps = {
   filterBy: string
 }
 
+const TableHead = ({ text }: { text: string }) => {
+  return (
```
Emmanueldmlr, line 69: why create another table head when you already created one?.

codiejay: There are inside independent files and almost too small to create dedicated files just for them.

PR #1010, opened 2022-12-14
Fixes high LCP and CLS readings

Royal-lobster:
# Changes
- fixes cite marks rendering multiline due to popup logic

# Screenshots
### Before fix [before JS load]
CleanShot 2022-12-14 at 15 08 41@2x

### After fix [before JS load]
CleanShot 2022-12-14 at 15 10 26@2x


Royal-lobster:
> looks good. lets link it to an issue related to lcp / improving seo score 👍🏻

Ohh forgot. thanks ! now its linked

PR #1014, opened 2022-12-15
Branding Link Fix

OleanjiKingCode:
# Fixed the broken branding link in iq wiki footer

_Description _

re-routed the link from "/static/branding" to "/branding"


Adebesin-Cell (approved):
Awesome 🚀

PR #1033, opened 2023-01-05
Create sitemap api

Emmanueldmlr:
Create sitemap api

fixes https://github.com/EveripediaNetwork/issues/issues/947

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -20,8 +20,6 @@ const config = {
   alchemyChain: process.env.NEXT_PUBLIC_ALCHEMY_CHAIN || 'maticmum',
   graphqlUrl:
     process.env.NEXT_PUBLIC_EP_API || 'https://api.dev.braindao.org/graphql',
-  apiUrl:
-    process.env.NEXT_PUBLIC_EP_API_DOMAIN || 'https://api.dev.braindao.org',
   editorGraphqlUrl:
     process.env.NEXT_PUBLIC_EDITOR_API ||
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -20,6 +20,8 @@ const config = {
   alchemyChain: process.env.NEXT_PUBLIC_ALCHEMY_CHAIN || 'maticmum',
   graphqlUrl:
     process.env.NEXT_PUBLIC_EP_API || 'https://api.dev.braindao.org/graphql',
+  apiUrl:
```
kesar, line 23: instead of adding extra endpoints i would suggest you strip /graphql from process.env.NEXT_PUBLIC_EP_API

src/pages/api/sitemap.ts
```diff
@@ -0,0 +1,13 @@
...
+import type { NextApiRequest, NextApiResponse } from 'next'
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<string>,
+) {
+  const url = `${config.apiUrl}/sitemap.xml`
+  const result = await fetch(url)
+  const response = await result.text()
+  res.setHeader('Cache-Control', 's-maxage=60')
```
kesar, line 11: dont u need to set headers for output xml ?

src/config/index.ts
Emmanueldmlr: done

src/pages/api/sitemap.ts
Emmanueldmlr: done

PR #1035, opened 2023-01-05
Quick footer year fix

OleanjiKingCode:
# Automated footer year

Before 
image

After

image

## Linked issues

 closes https://github.com/EveripediaNetwork/issues/issues/950

### DIFF
--- a/src/components/Layout/Footer/Footer.tsx
+++ b/src/components/Layout/Footer/Footer.tsx
@@ -36,6 +36,7 @@ const Footer = () => {
   const { t, i18n } = useTranslation()
   const spacing = useBreakpointValue({ base: 8, lg: 24 })
   const [lang, setLang] = useState<string>(languageData[0].value)
+  const thisYear = new Date().getFullYear()
 
   const handleLangChange = (userLang: string | string[]) => {
     if (isString(userLang)) {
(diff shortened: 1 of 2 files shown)


src/components/Layout/Footer/Footer.tsx
```diff
@@ -78,7 +78,10 @@ const Footer = () => {
         <SimpleGrid fontSize="sm" columns={{ base: 1, lg: 2 }}>
           <Stack align={{ base: 'center', lg: 'flex-start' }} flex="1">
             <Flex alignItems="center">
-              <Text py={3} pr={2}>{`${t('copyRight')}`}</Text>
+              <Text py={3} pr={2}>
+                {' '}
+                © {new Date().getFullYear()} {`${t('copyRight')}`}
```
Adebesin-Cell, line 83: I think you can store this in a const.

Adebesin-Cell (approved):
Awesome

src/components/Layout/Footer/Footer.tsx
OleanjiKingCode: Not sure it's needed in a var since it's only here we're using it. I'll do it tho

kesar: yup prob. not big deal

PR #1039, opened 2023-01-09
Fixes rank iteration

Adebesin-Cell:
Fixes rank iteration

Before:

issue

After:

Screen Shot 2023-01-09 at 3 25 24 PM

fixes https://github.com/EveripediaNetwork/issues/issues/948


Adebesin-Cell:
I’ll increase the font sizes then. 👍

Adebesin-Cell:
@Simplypheyie96 I've adjusted the font sizes.

Simplypheyie96 (changes requested):
The font size looks okay. 
Can't wait to finish the new ranking page. 
I think that would be so much more better.

also, why is the nft list not up to ten. It doesn't look good that way.
image

Adebesin-Cell:
> The font size looks okay. Can't wait to finish the new ranking page. I think that would be so much more better.
> 
> also, why is the nft list not up to ten. It doesn't look good that way. image

Ashaka said the backend is filtering NFt wikis without a 24h volume. For the first page autoglyphs is an example of such nft wiki.

Simplypheyie96:
> > The font size looks okay. Can't wait to finish the new ranking page. I think that would be so much more better.
> > also, why is the nft list not up to ten. It doesn't look good that way. image
> 
> Ashaka said the backend is filtering NFt wikis without a 24h volume. For the first page autoglyphs is an example of such nft wiki.

hmmnn.
you can go ahead and merge then.

Adebesin-Cell:
Yeah, I think we’ll review more on that 🙂.

src/components/Rank/InvalidRankCardItem.tsx
```diff
@@ -44,7 +44,7 @@ export const InvalidRankCardItem = ({
           >
             <Text
               color="inactiveText"
-              fontSize={{ md: 'xs', base: 'sm' }}
+              fontSize={{ base: 'xs' }}
```
Royal-lobster, line 47:
```suggestion
              fontSize="xs"
```
I think this is same as 'xs'

src/components/Rank/InvalidRankCardItem.tsx
```diff
@@ -57,8 +57,7 @@ export const InvalidRankCardItem = ({
               width="100%"
               justifyContent="end"
             >
-              {/* <AiFillCaretUp color={upIndicationIconColor} /> */}
-              <Text fontWeight="bold" fontSize={{ md: 'xs', base: 'xs' }}>
+              <Text fontWeight="bold" fontSize={{ base: 'xs' }}>
```
Royal-lobster, line 60:
```suggestion
              <Text fontWeight="bold" fontSize="xs" >
```
please check if it is same, if it is same we can directly use "xs".

Adebesin-Cell: Alright, I’ll look into i!t.

Adebesin-Cell:
@Simplypheyie96 I think you can check now, we have the 10 rankings listed out properly.

Softdev1 (approved):

Simplypheyie96:
> @Simplypheyie96 I think you can check now, we have the 10 rankings listed out properly.

good to go

PR #1045, opened 2023-01-11
Fix handling of null value keys from linked wikis

Royal-lobster:
closes https://github.com/EveripediaNetwork/issues/issues/957

# Changes
- added transformation to remove linkedWiki keys on create wiki page
- made removeLinkedWiki action to remove keys from linkedWikis object if it is empty and remove linkedWiki key from wiki object if it is empty object


Adebesin-Cell (approved):
I think it works now, we'll keep investigating if any errors comes up.

PR #1046, opened 2023-01-12
Factored out tokens calling the fetchTokenStats

Adebesin-Cell:
Factored out tokens calling the fetchTokenStats

Changes:

Call tokenStats only for coins.

Before:

Screen Shot 2023-01-12 at 11 19 31 AM

After:

Screen Shot 2023-01-12 at 11 20 58 AM

fixes https://github.com/EveripediaNetwork/issues/issues/958

### DIFF
--- a/src/components/Wiki/WikiPage/InsightComponents/NFTWidget.tsx
+++ b/src/components/Wiki/WikiPage/InsightComponents/NFTWidget.tsx
@@ -6,6 +6,7 @@ import {
   Heading,
   Image,
   Input,
+  Skeleton,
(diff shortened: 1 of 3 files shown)


Adebesin-Cell:
@Royal-lobster @OleanjiKingCode There was a bug yesterday rendering stats, so now we only call tokenStats Query for only tokens(coins). In later updaates, we'll look into rendering stats for NFTs too. 🚀

Royal-lobster (approved):

OleanjiKingCode (approved):

PR #1050, opened 2023-01-16
Adds wiki view

Emmanueldmlr:
# show views on the wiki page

add views to wikis show them if > 250

Result: 
image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/960

### DIFF
--- a/src/components/Editor/EditorModals/MediaModal.tsx
+++ b/src/components/Editor/EditorModals/MediaModal.tsx
@@ -214,7 +214,7 @@ const MediaModal = ({
                             handleSetType(media.id, e.target.value as MediaType)
                           }
                         >
-                          <option value={MediaType.DEFAULT}>Default</option>
+                          <option value={MediaType.GALLERY}>Default</option>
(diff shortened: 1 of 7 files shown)


kesar (approved):

src/components/Editor/EditorModals/MediaModal.tsx
```diff
@@ -89,7 +89,7 @@ const MediaModal = ({
         name,
         size: shortenBalance(fileSize),
         id,
-        type: MediaType.DEFAULT,
+        type: MediaType.GALLERY,
```
kesar, line 92: @Royal-lobster we were guessing that this is the right approach. If we missunderstood the DEFAULT - GALLERY thing let us know :pray

Royal-lobster: Yes ! we changed it DEFAULT to GALLERY because the backend is using the key of enum for graphql query. this is the right.

PR #1052, opened 2023-01-17
Fixes null error searching

Adebesin-Cell:
Fixes null error searching

Changes:

Added guard clause to response.

fixes https://github.com/EveripediaNetwork/issues/issues/965

### DIFF
--- a/src/components/Wiki/WikiAccordion/AccordionWidget.tsx
+++ b/src/components/Wiki/WikiAccordion/AccordionWidget.tsx
@@ -25,42 +25,59 @@ import { useENSData } from '@/hooks/useENSData'
 import { getUsername } from '@/utils/getUsername'
 import { WikiInsights } from '@/types/WikiInsightsDataType'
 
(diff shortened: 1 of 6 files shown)


Adebesin-Cell:
@kesar I think the error has been fixed here. In the tagById.wikis since the query doesn’t exist as a tag, the entire tagById is null from backend. Making the transform object null.wikis. I've added a guard clause, I think it should prevent the search error.

kesar:
why this pr has stuff related to linked wikis and nft widget?

Adebesin-Cell:
> why this pr has stuff related to linked wikis and nft widget?



> why this pr has stuff related to linked wikis and nft widget?

Ha, yeah. I updated linked wikis queries on wiki related queries. The NFT widget was a bug I noticed on the loading state so I added a fix for it, since it’s a small fix.

kesar:
ok, make sense 👍👍

Emmanueldmlr (approved):

PR #1066, opened 2023-01-19
Fixes rank page

Adebesin-Cell:
Fixes rank page

Changes:
Added tabled data for cryptocurrencies and NFTs.
Refactored common logic in rank homepage.
Added pagination buttons.
Added paginaation hook.
Added loading State using isFetching:

DONE:

- [x] Added tabs and data in homepage
- [x] Refactor the entire file.
- [x] Fix the new rank page design on iq.wiki/rank.
- [x] Add Paginations.
- [x] Add pagination buttons.
- [x] Refactor logic and component.
- [x] Fix numbering.
- [x] Add loading State.

Media:
localhost_4000_ (3)

localhost_4000_

fixes https://github.com/EveripediaNetwork/issues/issues/929


src/components/Rank/RankPagination.tsx
```diff
@@ -0,0 +1,88 @@
...
+        leftIcon={<AiOutlineArrowLeft />}
+        _hover={{ bg: 'transparent' }}
+        p="0"
+        bg="transparent"
+        _active={{ bg: 'transparent' }}
+        color="gray.700"
+        disabled={currentPage === 1}
+        onClick={() => {
+          onPrevious()
+        }}
```
Royal-lobster, line 60:
```suggestion
        onClick={onPrevious}
```

you can directly pass the function to onClick if no arguments are used

src/components/Rank/RankPagination.tsx
```diff
@@ -0,0 +1,88 @@
...
+        rightIcon={<AiOutlineArrowRight />}
+        _hover={{ bg: 'transparent' }}
+        p="0"
+        bg="transparent"
+        color="gray.700"
+        _active={{ bg: 'transparent' }}
+        disabled={currentPage === lastPage}
+        onClick={() => {
+          onNext()
+        }}
```
Royal-lobster, line 78:
```suggestion
        onClick={onNext}
```

src/pages/rank/index.tsx
```diff
@@ -1,37 +1,146 @@
...
+                onPageChange={page => setTokensOffset(page)}
+              >
+                <RankTableHead />
+                <Tbody>
+                  {tokensObject?.map((token, index) => {
+                    if (!token) {
+                      return <InvalidRankCardItem index={index} />
+                    }
+                    return <RankingItem index={index} item={token} />
+                  })}
```
Royal-lobster, line 108: Not something big but, this can be simplified to this: 

```jsx
tokensObject?.map((token, index) => token ?
<RankingItem index="{index}" item="{token}" />
:
<InvalidRankCardItem index="{index}" />
)
```

Simplypheyie96 (approved):
Great Implementation.
but we really need to get rid of the N/As and the coming soons.😪

src/components/Rank/RankCardItem.tsx
```diff
@@ -1,136 +1,124 @@
...
+          <Text color="rankingListText">NA</Text>
+        </Flex>
+      </Td>
+      <Td borderColor="rankingListBorder" fontWeight={500} fontSize="14px">
+        <Flex gap="1">
+          <Text color="rankingListText">
+            $
+            {item.nftMarketData
+              ? item.nftMarketData.market_cap_usd.toLocaleString()
+              : item.tokenMarketData.market_cap?.toLocaleString()}
```
Royal-lobster, line 65: These seems to be too long in JSX, you can extract them to a variable before returning so it reads better. something like this -> 

```jsx
const marketCapValue = `$
            {item.nftMarketData
              ? item.nftMarketData.market_cap_usd.toLocaleString()
              : item.tokenMarketData.market_cap?.toLocaleString()`
// ...
return (
//	...
	<Text color="rankingListText">{marketCapValue}</Text>
//	...
)
```

Royal-lobster:
@Adebesin-Cell is there no date founded information on api we are using ?

Adebesin-Cell:
> @Adebesin-Cell is there no date founded information on api we are using ?

Yeah, there's no way to get the date founded yet.

src/components/Rank/RankPagination.tsx
Adebesin-Cell: Ha, I'll fix that now.

Softdev1:
we would need a way to get all the date -founded @Adebesin-Cell and this should be called from the api direct

Softdev1:
we might to sync-up on the best way to go @Adebesin-Cell  and @Royal-lobster

Adebesin-Cell:
> we would need a way to get all the date -founded @Adebesin-Cell and this should be called from the api direct

Yeah, I think it'll be part of the metadata or there'll be a way to add it from the wiki page.

Softdev1 (commented):
Let's discuss on the best way to resolve the date-founded first

Softdev1:
Yeah, I was thinking we can create a tag for it and add it but @kesar  might have a better option for us to go

PR #1072, opened 2023-01-21
Makes blog generate on static revalidation

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/970

# Changes
- changes getServerSideProps to getStaticProps on blog page


kesar:
i think we would need some kind of revalidation

https://nextjs.org/docs/api-reference/data-fetching/get-static-paths#fallback-pages

perhaps revalidate each 5 min.

Royal-lobster:
> i think we would need some kind of revalidation
> 
> [nextjs.org/docs/api-reference/data-fetching/get-static-paths#fallback-pages](https://nextjs.org/docs/api-reference/data-fetching/get-static-paths#fallback-pages)
> 
> perhaps revalidate each 5 min.

Yes ! I am thinking we might do the same as we are doing for wiki pages, we can revalidate pages with web-hook, i am investigating if mirror has web hook feature on publish

Royal-lobster:
@kesar seems like there is no web hooks integration on mirror.xyz :( so added the 5 min revalidation instead.

Emmanueldmlr (approved):

Emmanueldmlr:
@Royal-lobster , still a WIP?

Royal-lobster:
> @Royal-lobster , still a WIP?

Nope 😅 forgot to remove from draft

kesar (approved):

PR #1083, opened 2023-01-25
Fixes cookie visit for first visit

Adebesin-Cell:
Fixes cookie visit for first visit

fixes https://github.com/EveripediaNetwork/issues/issues/969

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { Box, Flex } from '@chakra-ui/react'
 import { getPromotedWikis, wikiApi } from '@/services/wikis'
 import { store } from '@/store/store'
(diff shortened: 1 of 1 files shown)


src/pages/index.tsx
```diff
@@ -21,15 +21,53 @@ interface HomePageProps {
...
 }: HomePageProps) => {
+  // const userTime = JSON.parse(localStorage.getItem('FIRST_VISITED'))
+
+  const [userFirstVisit, setUserFirstVisit] = useState<
+    string | null | undefined
+  >(undefined)
+
+  const [showHero, setShowHero] = useState<boolean>(true)
+
+  const currentDate = useMemo(() => new Date(), [])
```
Royal-lobster, line 44: @Adebesin-Cell you can just use new Date() while setting the localstorage and read from there is needed elsewhere. Use memo should only be used for expensive operations.

Adebesin-Cell: Ha! I was trying to prevent the currentDate to re-render the component multiple times.

Emmanueldmlr (approved):

src/pages/index.tsx
Royal-lobster: You can initialise the currentDate directly in useEffect. Then it should't trigger the useEffect since it no longer will be a dependency 

CleanShot 2023-01-26 at 15 38 23@2x

Royal-lobster:
@Adebesin-Cell Also please test the page with google lighthouse. and make sure it is not effecting CLS score.

src/pages/index.tsx
Adebesin-Cell: Ha! I’ll fix that now.

Adebesin-Cell:
> @Adebesin-Cell Also please test the page with google lighthouse. and make sure it is not effecting CLS score.

I’ll inspect that too 👍.

Adebesin-Cell:
@kesar @Simplypheyie96 @Royal-lobster @Softdev1 

Now:
Screen Shot 2023-01-30 at 3 45 08 PM

src/utils/getFirstUserVisit.ts
```diff
@@ -0,0 +1,24 @@
+const TIME_LIMIT = 60
+
+export const userFirstVisit = () => {
```
kesar, line 3: not needed. if we need it again we can check the commits here

src/pages/index.tsx
```diff
@@ -41,20 +56,20 @@ export const Index = ({
   rankings,
 }: HomePageProps) => {
   return (
-    <Flex direction="column" mx="auto" w="full" pt={{ base: 6, lg: 20 }}>
-      <Hero wiki={promotedWikis && promotedWikis[0]} />
+    <Flex direction="column" mx="auto" w="full" pt={{ base: 6, lg: 12 }}>
+      <HeroAfterFirstVisit />
```
kesar, line 60: name is wrong now

src/utils/getFirstUserVisit.ts
Adebesin-Cell: Alright sir, I'll remove it now.

kesar (approved):

PR #1095, opened 2023-01-30
little fix on the founders and blockchain tags

OleanjiKingCode:
# little fix on the founders and blockchain tags

Before 
image

After

image


Adebesin-Cell (approved):
🚀

PR #1096, opened 2023-01-30
fixes incorrect numbering on rank page

Adebesin-Cell:
fixes incorrect numbering on rank page

Before:

https://user-images.githubusercontent.com/32770520/215434554-2a9bd471-1f90-42c9-a2eb-0a563011a5a6.png

Now:

localhost_4000_rank (1)

localhost_4000_rank (2)

fixes https://github.com/EveripediaNetwork/issues/issues/991

### DIFF
--- a/src/pages/rank/index.tsx
+++ b/src/pages/rank/index.tsx
@@ -41,8 +41,8 @@ const Rank = ({
   const [nftCount, setNftCount] = useState<number>(0)
 
   useEffect(() => {
-    setTokenCount(tokensOffset * 10 - 10)
-    setNftCount(nftOffset * 10 - 10)
+    setTokenCount(tokensOffset * 20 - 20)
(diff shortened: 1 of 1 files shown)


src/pages/rank/index.tsx
```diff
@@ -41,8 +41,8 @@ const Rank = ({
   const [nftCount, setNftCount] = useState<number>(0)
 
   useEffect(() => {
-    setTokenCount(tokensOffset * 10 - 10)
-    setNftCount(nftOffset * 10 - 10)
+    setTokenCount(tokensOffset * 20 - 20)
+    setNftCount(nftOffset * 20 - 20)
```
kesar, line 45: this should be LISTING_LIMITS to avoid using magic numbers

Adebesin-Cell: Okay @kesar

PR #1107, opened 2023-02-01
Fixes wiki page runtime errors

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/995

# Changes
- fixes navbar errors due to ssr. (https://github.com/chakra-ui/chakra-ui/issues/3020)
- fixes key errors due to no key prop. 
- fixes div in span error by replacing div with span.

# Screenshots

CleanShot 2023-02-04 at 21 31 45@2x


src/components/Elements/LinkElements/LinkWrapper.tsx
```diff
@@ -4,12 +4,18 @@ import NextLink, { LinkProps } from 'next/link'
 
 type ChakraLinkAndNextProps = ChakraLinkProps & LinkProps
 
+const POLYGON_MUMBAI_ERROR_LINK = 'https://mumbai.polygonscan.com//tx/'
```
kesar, line 7: we shouldnt leak test env code into prod. I guess there must be a root problem that leads to this.

also this is very inefficient, take in consideration that we would be checking for every single link wrapper invocations to see if it includes a tx from mumbai that will only happens on dev and in very few situations.

Adebesin-Cell: Ha! Noted @kesar. 🙏

kesar (approved):

PR #1112, opened 2023-02-02
Rearranging utils files

OleanjiKingCode:
# Rearranging utils files

_Rearranging utils files for best pratice_

Before
image

After
image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1008


Royal-lobster (commented):
@OleanjiKingCode Great work this is looking better 👏 some more points
- the useInfiniteData can be moved to hooks folder
- the `PasteListener.ts` and `create-wiki.ts` can be in one folder like createWikiUtils. this can be useful when we refactor the `create-wiki.ts` file since it is too big.
- we can also have one folder for all blockchain related utils with some files inside dataFetching folder and `postSignature.ts`
- the `networkMap.ts` can be moved to data folder since it is just js object.

CleanShot 2023-02-02 at 13 26 21@2x

kesar:
good start. make sure folders and files follow same name convention.

some folders start w capital letter, other dont

some files are with dashes, others start w capital letter and other dont. reach consensus on that based in best practices 👍

Adebesin-Cell (commented):
Screen Shot 2023-02-03 at 12 44 37 PM

You can store this in a folder named DataTranform | Conversion since they work on getting the raw data into a proper one.

Royal-lobster:
Almost all functions in these files are taking some kind of wiki object and giving back the data. may be we can club them in one a folder `WikiUtils`. 

CleanShot 2023-02-03 at 18 27 30@2x
CleanShot 2023-02-03 at 18 27 05@2x
CleanShot 2023-02-03 at 18 26 46@2x
CleanShot 2023-02-03 at 18 26 37@2x


- May be its simpler to maintain the folders based on features like `WikiUtils`,`CreateWikiUtils`, `BlogUtils`, `WalletUtils`, `AdminUtils`, `ApiUtils` ? not every feature can have its own folder. i think those were only folders we might need to store almost all util files. if one folder becomes too bit, we can again split it to sub features which takes more util files.
- The entire files in DataConversion folder can be clubed to one file named `textUtils.ts` since all of them are short functions. combining them won't make the file bigger than 50 lines.

PR #1123, opened 2023-02-03
Fixes and Refactors wiki share modal

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/1018

# Changes
- changes share modal design to be more compact on mobile
- makes use of common modal element instead of using directly from chakra ui

# Screenshots
Frame 397

### DIFF
--- a/src/components/Wiki/WikiPage/CustomModals/ShareWikiModal.tsx
+++ b/src/components/Wiki/WikiPage/CustomModals/ShareWikiModal.tsx
@@ -1,17 +1,15 @@
 import React from 'react'
 import {
   ModalProps,
-  Modal,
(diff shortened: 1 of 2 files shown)


kesar (approved):

kesar:
great refactoring 👍🏻 btw not needed to fix here, but do we track dialog open by any chance? at some point i would like to know how many people try to share (not big deal)

Royal-lobster:
> great refactoring 👍🏻 btw not needed to fix here, but do we track dialog open by any chance? at some point i would like to know how many people try to share (not big deal)

Good point ! We are not tracking currently. I will add the code to my current working PR.

PR #1139, opened 2023-02-07
Fixes Careers Page

Adebesin-Cell:
Fixes Careers Page

links:
https://iq-wiki-git-fixes-careers-page-prediqt.vercel.app/careers

Screenshots:
localhost_4000_careers

fixes https://github.com/EveripediaNetwork/issues/issues/1024


src/data/CareersData.ts
```diff
@@ -0,0 +1,18 @@
...
+  link: string
+  location: string
+}
+
+export const AllCareers: CareersDatatype[] = [
+  {
+    id: 1,
+    title: 'AI developer',
+    description:
+      'Are you an AI Developer with an insatiable appetite for a challenge? Then this job opportunity is for you! We are looking for an ambitious and creative AI Developer to join our team and help solve some of the most complex problems in the tech industry. You will have the opportunity to work with the most cutting edge technologies and develop innovative solutions that have the potential to make a real impact on the world. If you’re looking to take your AI development career to the next level, then this is the job for you! So, c’mon – let’s get coding and make some magic!',
```
kesar, line 14: make sure it support line jumps

src/data/CareersData.ts
```diff
@@ -0,0 +1,18 @@
...
+  id: number
+  title: string
+  description: string
+  link: string
+  location: string
+}
+
+export const AllCareers: CareersDatatype[] = [
+  {
+    id: 1,
```
kesar, line 11: i dont think we need ids, we can use key in the react loop

Adebesin-Cell: Noted @kesar. I'll remove it.

src/data/CareersData.ts
Adebesin-Cell: I'll investigate that @kesar.

kesar:
do we need a 1900x900 images for the hero? and dont we have those heros somewhere already?

also yarn.lock is updated without any new package added, so this is most likely something coming from other previous PR. probably worth revert yarn

Adebesin-Cell:
> do we need a 1900x900 images for the hero? and dont we have those heros somewhere already?
> 
> we have similar / same brainies bg (that i really love :P) in homepage and as header in profile page if user didnt setup one. I think we should try to reuse them if possible
> 
> also yarn.lock is updated without any new package added, so this is most likely something coming from other previous PR. probably worth revert yarn

Okay @kesar, I'll add the fix now. I'll also revert yarn.

Adebesin-Cell:
> code looks great now. kudos! the only missing things are linting & line jumping in job texts 👍🏻

Thanks @kesar. I'll look into how to implement line jumps and fix the lints.

Softdev1:
Well done @Adebesin-Cell  looks great

Simplypheyie96 (changes requested):
- The card in dark mode should be in gray700
image

- we're hiring tag is missing in footer
image

Brainies background not visible in dark mode

Simplypheyie96 (approved):

Royal-lobster (approved):

PR #1170, opened 2023-02-20
Content feedback

OleanjiKingCode:
# Wiki Feedback for Users

## How should this be tested?

UI

- ## Question
image
image

- ## Feedback

image
image

## Linked issues
closes https://github.com/EveripediaNetwork/issues/issues/1007


Simplypheyie96 (approved):

Softdev1 (approved):
awesome

src/components/Wiki/WikiPage/InsightComponents/ContentFeedback.tsx
```diff
@@ -0,0 +1,83 @@
...
+import WikiAccordion from '../../WikiAccordion'
+
+const ContentFeedback = ({
+  feedback,
+  wikiId,
+  userId,
+}: {
+  feedback: boolean
+  wikiId: string
+  userId: string | undefined
```
AshakaE, line 15: - [ ] I suggest we don't send undefined for the users, feedback should only be possible for logged in users

AshakaE (changes requested):

src/components/Wiki/WikiPage/InsightComponents/ContentFeedback.tsx
OleanjiKingCode: Alright

kesar: we barely have registered users, it would defeat the purpose of knowing if content is good or bad.

we should allow anonymous users until we have enough userbase

OleanjiKingCode: @kesar using without db that makes users (anonymous or not ) like more than once

AshakaE: I think we'll need to store some information about the feedback to prevent spamming

OleanjiKingCode: yh

AshakaE (approved):

src/components/Wiki/WikiPage/InsightComponents/ContentFeedback.tsx
kesar: api should do something similar to what we have for views and prevent that

AshakaE: thought as much, I just wanted to be sure 👍

PR #1183, opened 2023-02-22
Feedback widget quick fix

OleanjiKingCode:
# Quick fix on the feedback widget


src/types/admin.ts
```diff
@@ -10,7 +10,7 @@ export type EditorsType = {
 
 export type ContentFeedbackArgs = {
   wikiId: string
-  userId: string
+  userId: string | undefined
```
Adebesin-Cell, line 13:
```suggestion
  userId?: string
```

You can use it like this.

PR #1185, opened 2023-02-22
fixing admin rework

Adebesin-Cell:
fixing admin rework
closes https://github.com/EveripediaNetwork/issues/issues/1037


src/components/Admin/WikiEditorInsight/InsightTableWikiEditors.tsx
```diff
@@ -18,67 +18,60 @@ import {
...
+//     wikiId: string
+//     datetime: string
+//     ipfs: string
+//     content: { title: string; images: { id: string }[] }[]
+//   }
+//   editorAvatar: string
+//   latestActivity: string
+//   editorAddress: string
+//   active: boolean
+// }
```
Royal-lobster, line 60: @Adebesin-Cell remove this if not used

```suggestion
```

Royal-lobster (approved):

Royal-lobster:
Not need in this pr, but i think we need a loading state when user clicks on next under editors table / wiki table. there is a delay on fetching which should fill in the time so user don't think table didn't change after clicking next

Adebesin-Cell:
Yeah, I’ll ping @Simplypheyie96 for that.

PR #1204, opened 2023-02-27
fixes category title transform function

Adebesin-Cell:
fixes category title transform function


Royal-lobster (approved):

src/utils/DataTransform/changeCategoryTitle.ts
```diff
@@ -0,0 +1,25 @@
+export const manualTranformCategoryTitle = (categoryType: string) => {
```
Royal-lobster, line 1: lol i think the title can be TransformCategoryTitle but its fine 😅

Adebesin-Cell: 😅, Alright. I'll change that.

PR #1206, opened 2023-02-27
fixes content feedback iteration

Adebesin-Cell:
fixes content feedback iteration

fixes https://github.com/EveripediaNetwork/issues/issues/1075

### DIFF
--- a/src/components/Wiki/WikiAccordion/index.tsx
+++ b/src/components/Wiki/WikiAccordion/index.tsx
@@ -15,22 +15,32 @@ interface AccordionProps {
   withNoDarkBg?: boolean
   collapsed?: { base: boolean; xl: boolean }
   children: React.ReactNode
+  normal?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


src/components/Wiki/WikiAccordion/index.tsx
```diff
@@ -15,22 +15,32 @@ interface AccordionProps {
   withNoDarkBg?: boolean
   collapsed?: { base: boolean; xl: boolean }
   children: React.ReactNode
+  normal?: boolean
```
kesar, line 18: whats normal?

Adebesin-Cell: A default state prop, I renamed it now 🙏. If true, the accordion is collapsed by default and if false/undefined, the accordion is uncollapsed.

OleanjiKingCode (approved):

Royal-lobster (approved):

PR #1211, opened 2023-02-28
Event feature

Adebesin-Cell:
Event feature

localhost_4000_create-wiki_slug=bored-ape-yacht-club(Nest Hub Max) (1)

closes https://github.com/EveripediaNetwork/issues/issues/986


Softdev1:
Please review the title section it seems it should be read-only cos one can't type there @Adebesin-Cell

Adebesin-Cell:
> Please review the title section it seems it should be read-only cos one can't type there @Adebesin-Cell

Hi @Softdev1, for the first events, the check is to have a Date of Creation or Date of Birth as the first title of the event depending on the wiki category. So it's readonly on first inout, but editable after the first inputs 👍

Softdev1:
I will have it tested again and revert back to you @Adebesin-Cell

Adebesin-Cell:
> I will have it tested again and revert back to you @Adebesin-Cell

Sweet! 🙂

Simplypheyie96 (changes requested):
-This should have a default text guide in it, like (enter event title), it will be confusing to see this
image

- let's use another date selector that would be more user-friendly. 

besides that, everything looks good

Royal-lobster:
CleanShot 2023-03-07 at 14 20 01@2x

@Adebesin-Cell make this read only as you said previously and also add a tooltip with sentence saying like it cannot be changed

Adebesin-Cell:
> CleanShot 2023-03-07 at 14 20 01@2x
> 
> @Adebesin-Cell make this read only as you said previously and also add a tooltip with sentence saying like it cannot be changed

Yeah, it's read only when the wiki doesn't have an event.

Royal-lobster:
image

We also have to update the blocks changed when every events are updated. and also events page should be revalidated on backend as well

Softdev1:
We might still need to wait and properly test this PR again so it will only get merged after it is properly tested and confirmed.
@Adebesin-Cell  you would need to sync up with @AshakaE  cos something which I gave you feedback are still not solved.

Softdev1:
Also, when a change is made on any event, it should be captured in the wiki history page @AshakaE , currently, this is not happening.

src/pages/wiki/[slug]/events.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </SimpleGrid>
+          )}
+        </SimpleGrid>
+      )
+    } else {
+      eventContent = <NoEventView />
+    }
+  } else {
+    eventContent = <NoEventView />
+  }
```
Comment (Royal-lobster) [inline src/pages/wiki/[slug]/events.tsx:73]: looks like duplicate code, i think this can be fixed

src/store/slices/wiki.slice.ts
```diff
@@ -254,6 +257,94 @@ const wikiSlice = createSlice({
...
     },
+    addEvent(state, action) {
+      const { title, description, type, date, link } =
+        action.payload as BaseEvents
+
+      let index = -1
+
+      if (state.events) {
+        index = state.events.findIndex(event => event.date === date)
+      }
```
Royal-lobster, line 268:
```suggestion
      const index = state.events ? state.events?.findIndex(e => e.date === date) : -1
```

Royal-lobster (approved):
Looks great ! we are good to merge after fixes !

src/pages/wiki/[slug]/events.tsx
Comment (Adebesin-Cell) [reply src/pages/wiki/[slug]/events.tsx:73]: I'll fix this.

Comment (Adebesin-Cell) [reply src/pages/wiki/[slug]/events.tsx:73]: I think if the wiki doesn't exist we should use a no-wiki component.

Comment (Royal-lobster) [reply src/pages/wiki/[slug]/events.tsx:73]: i believe it should simply go to 404. or may be the wiki page since it has create wiki action

Royal-lobster (approved):

Adebesin-Cell:
@Royal-lobster @Softdev1 Good to merge?

Royal-lobster:
Yep lets get this merged !

PR #1235, opened 2023-03-09
fixes link button

Adebesin-Cell:
fixes link button


Adebesin-Cell:
@kesar I think this was why the events didn't display. `wiki.events && wiki.events.length >= 1`

kesar:
makes sense 👍🏻

PR #1260, opened 2023-04-06
Fix account rounting

Emmanueldmlr:
# Improper Routing of the setting page

_As a Developer, I need to ensure that the settings page is properly routed, so that users can easily navigate to and modify their preferences.

fixes https://github.com/EveripediaNetwork/issues/issues/1189


Emmanueldmlr:
@Softdev1, we need to test this update aggressively just to be sure nothing breaks.

Softdev1:
sure we would ensure nothing breaks via careful and precise testing @Emmanueldmlr

kesar:
why u didnt move the /settings path that has only 1 or 2 links ? we can't replace route of accounts that will affect to SEO

Emmanueldmlr:
oops. I was much more particular about the URL. I will rework it now.

kesar:
yeah i think that would work better 👍🏻 👍🏻  thx!

Softdev1 (approved):
Tested and it is working fine.

PR #1265, opened 2023-04-11
Add rome

Emmanueldmlr:
# add Rome to ep-ui

_add Rome to ep-ui. do it after merging the 2 pending PR so we dont run into a big amount of conflicts

fixes https://github.com/EveripediaNetwork/issues/issues/1193


Emmanueldmlr:
I think we are good to merge

Emmanueldmlr:
We noticed that the links under the Community Hub are not routed to the right social media pages, I realized that it's like that in production. We will create a new issue to handle that.

PR #1269, opened 2023-04-12
when user isnt logged in feedback

OleanjiKingCode:
# A quick fix for the iq wiki content feedback to handle when users arent logged in and they provide a fedback


AshakaE (approved):
Perfect 👍

PR #1282, opened 2023-04-17
Wiki views tracker

OleanjiKingCode:
# Wiki Views Tracker and Admin UI update

Before :

image

image

Now:

image

image

image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1185

### DIFF
--- a/src/components/Admin/GraphHeads.tsx
+++ b/src/components/Admin/GraphHeads.tsx
@@ -3,7 +3,12 @@ import { Th, Text } from '@chakra-ui/react'
 
 export const TableHead = ({ text }: { text: string }) => {
   return (
-    <Th color="primaryGray" textTransform="capitalize" fontWeight="normal">
+    <Th
+      color="primaryGray"
+      _dark={{ color: 'white' }}
+      textTransform="capitalize"
+      fontWeight="normal"
+    >
       <Text fontWeight="bold">{text}</Text>
     </Th>
   )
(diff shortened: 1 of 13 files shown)


Simplypheyie96 (changes requested):
according to the prompt on the issue page, the graph has to be in dates, so it can correlate with the table. implement according to the design.
image

OleanjiKingCode:
Yes that's dummy data for now until the data is gotten form the backend

Simplypheyie96:
> Yes that's dummy data for now until the data is gotten form the backend

i am talking about the days below, mon, tues,wed,
In the design, you'll see it as day/month (dates). not just name of days
image

kesar:
for simplicity API just returns days, so no need the dropdown. also just make the chart to be last 365 days (or less if we have less data, i think rn we might have only 6 months back)

Simplypheyie96:
> for simplicity API just returns days, so no need the dropdown. also just make the chart to be last 365 days (or less if we have less data, i think rn we might have only 6 months back)

@OleanjiKingCode that means the graph will be this way without the daily dropdown we had going before
image

OleanjiKingCode:
image

image

image

OleanjiKingCode:
@kesar now thats there visits count, how does the user data with editors and visitors work

Adebesin-Cell (commented):
Screen Shot 2023-04-20 at 2 26 39 PM

Weirdly, no data is showing for editors. :))

OleanjiKingCode:
Yeah, true but that's not part of my pr

Adebesin-Cell:
> Yeah, true but that's not part of my pr

Data shows on search tho 🙂. Look into it and see what's the cause, maybe if it's something huge you can make a new PR.
I think it's from this PR @AshakaE, look at the backend maybe.

## Main branch currently.
Screen Shot 2023-04-20 at 2 38 56 PM

Royal-lobster (approved):

PR #1298, opened 2023-04-26
Structuring env vars

OleanjiKingCode:
# Structuring env vars

The structure can be seen in the .env.sample file

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1248


Emmanueldmlr:
@OleanjiKingCode , why do we have googleAnalyticsScript twice?
image

src/env.mjs
```diff
@@ -0,0 +1,115 @@
...
+  REVALIDATE_PAGES_SECRET: z.string(),
+  OPENAI_API_KEY: z.string().optional(),
+  PINATA_KEY: z.string().optional(),
+  PINATA_SECRET: z.string().optional(),
+})
+
+const client = z.object({
+  NEXT_PUBLIC_IS_PRODUCTION: z.string(),
+  NEXT_PUBLIC_INFURA_ID: z.string(),
+  NEXT_PUBLIC_ALCHEMY_API_KEY: z.string(),
```
Emmanueldmlr, line 14: I think this should be moved to the server, we also don't want to expose this I guess.

src/env.mjs
```diff
@@ -0,0 +1,115 @@
...
+  NEXT_PUBLIC_PINATA_GATEWAY_BASE_URL: z.string().url(),
+  NEXT_PUBLIC_MAGIC_LINK_API_KEY: z.string(),
```
Emmanueldmlr, line 16: We should move this to the server too.

src/env.mjs
```diff
@@ -0,0 +1,115 @@
...
+  NEXT_PUBLIC_ENS_RPC: z.string().url(),
+  NEXT_PUBLIC_GOOGLE_ANALYTICS: z.string(),
```
Emmanueldmlr, line 18: move this to the server too.

OleanjiKingCode:
> @OleanjiKingCode , why do we have googleAnalyticsScript twice? image

Not sure about that i just changed the env vars used there, ill check it though

src/env.mjs
kesar: i think googel analytics need to be on the client

src/env.mjs
OleanjiKingCode: expose??

kesar: not sure if it will work if you move it to backend.. do testing if you do these changes

Royal-lobster (approved):

PR #1303, opened 2023-04-27
Fix wallet bug

Emmanueldmlr:
# Wallet Bug

_ As a user, I would want to view the both total values of all tokens and also the breakdown of each token in my wallet. Currently, on the wallet, a User can't see the total values of all the tokens in the wallet, this is because it is displaying NAN USD.

result 
image

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/1245


Royal-lobster (approved):

kesar:
hiiq should have iq value imo. hopefully its not 0 :P

PR #1312, opened 2023-05-03
404 error pages

OleanjiKingCode:
# Error pages for both Tags and Wikis

When users goes to a non existing tag / wiki page we are to show a 404 error page

Before: 
image
image

Now: 

image

image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1289,
closes https://github.com/EveripediaNetwork/issues/issues/1290


src/pages/wiki/[slug]/index.tsx
```diff
@@ -69,9 +69,9 @@ export const getStaticProps: GetStaticProps = async (context) => {
 
   // TODO: probably can be async in the components
   const { data } = await store.dispatch(getWikiCreatorAndEditor.initiate(slug))
-
+  const wikiData = wiki ? { ...wiki, ...data } : null
```
Comment (Adebesin-Cell) [inline src/pages/wiki/[slug]/index.tsx:72]: I think we should handle the whole redirect in this component, returning ```null``` and handling in the ```WikiMarkup``` component still makes the SEO contents pass as the not found wiki slug.

Adebesin-Cell (changes requested):

src/pages/tags/[tag].tsx
```diff
@@ -110,13 +113,35 @@ const TagPage: NextPage<TagPageProps> = ({ tagId, wikis }: TagPageProps) => {
...
                 </Center>
               )}
-            </>
-          ) : (
-            <Box textAlign="center" py={10} px={6}>
-              <Text fontSize="lg" mt={3} mb={3}>
+            </Box>
+          </>
+        ) : (
+          <Flex
```
Comment (Adebesin-Cell) [inline src/pages/tags/[tag].tsx:119]: I think we can also redirect the tag if there's no wiki found on it in the getServerSideProps.

Adebesin-Cell (approved):
Looks good!

PR #1313, opened 2023-05-03
improve wiki load speed on profile page

Tolu-Mals:
# Improve wiki load speed on profile page

closes https://github.com/EveripediaNetwork/issues/issues/1251

The problem was that the profile page was only using ssr for the profile data, but not the wikis. I used ssr to fetch the wikis as well. The wikis load much faster now ⚡️

## How should this be tested?
While logged in,
- Hover on the profile icon
- Click 'profile'

The wikis should not take 4s to load, they should show up immediately.

## Video Demo
https://user-images.githubusercontent.com/30352484/236177485-a55c456a-3b26-440d-afca-b33bfbedfad8.mov


Royal-lobster (approved):

Emmanueldmlr (approved):

Tolu-Mals:
> great pr 👍🏻

Thank you ✨

PR #1333, opened 2023-05-18
updates wiki initial message

Emmanueldmlr:
#updates wiki initial message

It takes a long period of time before a wiki gets published, and the user would need to wait for about 5 minutes since it would need to be signed/approved on the blockchain. we need to review the time for this to happen and show a pop-up message for the delay.

## Linked issues

fixes https://github.com/EveripediaNetwork/issues/issues/1329


Emmanueldmlr:
image

Simplypheyie96 (approved):

PR #1340, opened 2023-05-22
removes us career page

Emmanueldmlr:
# Removes us career page

Based on the instruction of Navin, we would need to entirely remove page the Us-career page from the IQ.wiki page.

Here is the link to the page: https://iq.wiki/us-careers

fixes https://github.com/EveripediaNetwork/issues/issues/1375


Softdev1 (approved):
Nice

PR #1405, opened 2023-07-09
Update magic connector

Emmanueldmlr:
# Update magic connector

_ We moved the magic package repository to magic recently
( https://www.npmjs.com/package/@magiclabs/wagmi-connector ).

it would be great if we can swap ours with this on iq wiki/dashboards

fixes https://github.com/EveripediaNetwork/issues/issues/1529


Royal-lobster:
Awesome ! is there any type errors after update ?

Emmanueldmlr:
> Awesome ! is there any type errors after update ?

No actually.

PR #1408, opened 2023-07-11
Inclusion of a copy icon on the wallet drawer's page

Emmanueldmlr:
# Inclusion of a copy icon on the wallet drawer's page

- [x] We would want to make some adjustments to the drawer's session of the IQ.wiki page, on the right side where the address is, let's get a button to copy the address to the clipboard
- [x] Also, on the drawer page, let's have a link to the profile page of the user's page instead of having the meta mask

fixes https://github.com/EveripediaNetwork/issues/issues/1527

### DIFF
--- a/src/components/Icons/CopyIcon.tsx
+++ b/src/components/Icons/CopyIcon.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+import { Icon, IconProps, useColorModeValue } from '@chakra-ui/react'
+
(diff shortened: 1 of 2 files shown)


Softdev1 (changes requested):
SCR-20230711-ilcy
the metamask on that drawer should be gone

Emmanueldmlr:
@Softdev1 , fixed now.

Softdev1 (approved):

PR #1414, opened 2023-08-04
fix blinking transition 

Tolu-Mals:
# Summary

Fixes the theme glitch when navigating to a page

## How should this be tested?

1. Load any page
2. Refresh the page

## Linked issues

Related https://github.com/EveripediaNetwork/issues/issues/1549

PR #1416, opened 2023-08-04
fix theme glitch with next-themes

Tolu-Mals:
# Summary

Attemps to fix the theme glitch issue using next-themes as discussed [here](https://github.com/pacocoursey/next-themes/issues/59#issuecomment-1060604644)

## How should this be tested?

1. Reload a page

## Notes or observations

This issue happens only on prod, so we can only know if this works for sure after merging to main

## Linked issues

Related #1414


Adebesin-Cell (approved):
Looks fine on dev.

Softdev1:
The main issue is not on the dev, the main issue is on production, so it is until we merge, we can confirm its state, unfortunately.

Tolu-Mals:
Issue wasn't fixed. I'm reverting because I installed an extra pacakage (next-themes)

PR #1420, opened 2023-08-10
Fix theme glitch

Tolu-Mals:
# Summary
Closes https://github.com/EveripediaNetwork/issues/issues/1549. 

I think this pr should finally fix this bug. The approach is to use a script that blocks the UI while determining the correct theme and then set's the `data-theme` property on the html element (which sets the app's theme). There's the flash of a blank white page before the UI shows, but I don't think that's an issue.


Tolu-Mals:
still has some issues lol, working on it

Tolu-Mals:
Everything should work fine now

Emmanueldmlr (approved):

Tolu-Mals:
still didn't work 😬

Tolu-Mals:
I think the script isn't running on production, I'm not sure why

PR #1435, opened 2023-08-16
Attempt to fix color mode flash

Royal-lobster:
# Changes
- tiny attempt to fix color mode flash


Royal-lobster:
did not work :/

PR #1496, opened 2023-09-21
Admin page fix

Aliiiu:
# Metric state change before render

This PR fixes the issue with the percentage state on the wiki meta data card. 

## Notes or observations

This issue is only visible on page load or reload. the percentage shows NAN as the value

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/


src/components/Admin/WikiDetailsCards.tsx
```diff
@@ -80,9 +80,9 @@ export const WikiDetailsCards = ({
               {weeklyValue} this week
             </Text>
           </VStack>
-          <CircularProgress value={percentage} color={color} size="45px">
+          <CircularProgress value={percentage || 0} color={color} size="45px">
```
Adebesin-Cell, line 83: ``CircularProgressValue = percentage || 0 ``, kindly replace the ``percentage || 0`` anywhere it's used.

```suggestion
          <CircularProgress value={CircularProgressValue} color={color} size="45px">
```

Adebesin-Cell (approved):
great work 👏. You can merge after the suggestion.

PR #1498, opened 2023-09-27
Feat add AI tokens

SidharthK2:
# Add `AI Tokens` tab to rankings table on landing and rank page.

Adds AI token table to ranking table. Existing token data is filtered and rendered in AI tokens tab sorted by market cap.

WIP screenshot

image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1799

### DIFF
--- a/src/components/Landing/RankingList.tsx
+++ b/src/components/Landing/RankingList.tsx
@@ -51,8 +51,8 @@ const RankingList = ({ rankings, listingLimit }: RankingListProps) => {
     setAiTokenItems(
       sortByMarketCap(
         'descending',
-        TokensListing.filter(
-          (item) => item?.tokenMarketData?.isAiToken === true,
+        TokensListing.filter((item) =>
+          item.tags.some((tag) => tag.id === 'AI'),
         ),
       ),
     )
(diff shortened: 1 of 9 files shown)


Aliiiu (approved):

Aliiiu:
> Details

since we have no data why are we having pagination

Aliiiu (approved):

Simplypheyie96 (changes requested):
image

doesn't look good on mobile, you can reduce the size of the icons, the text and the spacing between the tabs @SidharthK2

Simplypheyie96:
> image
> 
> doesn't look good on mobile, you can reduce the size of the icons, the text and the spacing between the tabs @SidharthK2 also, we can make it the default size but the tab scrollable. just like the table so as to not squeeze the tab

image
something like this. so the tab is scrollable too @SidharthK2

Simplypheyie96:
> > > image
> > > doesn't look good on mobile, you can reduce the size of the icons, the text and the spacing between the tabs @SidharthK2 also, we can make it the default size but the tab scrollable. just like the table so as to not squeeze the tab
> > 
> > 
> > image something like this. so the tab is scrollable too @SidharthK2
> 
> @Simplypheyie96 so keep the table as is, and make it side-scrollable?

let's do it that way. so we won't have to squeeze anything

Simplypheyie96 (approved):

Softdev1 (approved):

AshakaE (approved):

PR #1505, opened 2023-10-05
Youtube link regex fix

Aliiiu:
# Youtube link regex fix

This PR fixes the issue with the input field not accepting '@' as part of the special characters that should be accepted for links.

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1844

### DIFF
--- a/src/data/WikiLinks.ts
+++ b/src/data/WikiLinks.ts
@@ -83,7 +83,10 @@ export const LINK_OPTIONS = [
     type: LinkType.SOCIAL,
     label: 'Youtube',
     icon: AiFillYoutube,
-    tests: [/https:\/\/(www.)?youtube.com\/\w+/],
+    tests: [
+      /https:\/\/(www.)?youtube.com\/\w+/,
+      /https:\/\/www\.youtube\.com\/(@\w+)\b/,
+    ],
   },
   {
     id: CommonMetaIds.FACEBOOK_PROFILE,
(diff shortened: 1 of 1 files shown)


Softdev1 (approved):

Adebesin-Cell (approved):
Looks good.

src/data/WikiLinks.ts
```diff
@@ -83,7 +83,10 @@ export const LINK_OPTIONS = [
     type: LinkType.SOCIAL,
     label: 'Youtube',
     icon: AiFillYoutube,
-    tests: [/https:\/\/(www.)?youtube.com\/\w+/],
+    tests: [
+      /https:\/\/(www.)?youtube.com\/\w+/,
+      /https:\/\/www\.youtube\.com\/(@\w+)\b/,
```
Adebesin-Cell, line 88: Hmm, try to handle if www isn't present

Royal-lobster (approved):

src/data/WikiLinks.ts
Royal-lobster: may be this would work 

/https:\/\/(www.)?youtube.com\/[\w@]+/

Adebesin-Cell: Combining regex logic sometimes can be buggy, though not sure for simple ones like this.

Aliiiu: Dope, it works for both

PR #1519, opened 2023-10-12
Fix wiki not being linked

SidharthK2:
# Fixes blockchain wiki not being linked in wiki metadata editor

- Used a getWikiByTitle function to fetch the selected linked wiki, instead of using the wiki title.

image

This part needs to remain same, as the `value` prop being different was causing the bnb chain bug according to my research.

### DIFF
--- a/src/components/CreateWiki/EditorModals/EditWikiDetailsModal/LinkedWikisInput.tsx
+++ b/src/components/CreateWiki/EditorModals/EditWikiDetailsModal/LinkedWikisInput.tsx
@@ -96,12 +96,11 @@ const LinkedWikisInput = ({ wiki }: { wiki: Wiki }) => {
     wiki.linkedWikis &&
     Object.values(wiki.linkedWikis).some((wikis) => wikis.length > 0)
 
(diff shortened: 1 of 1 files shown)


Royal-lobster:
@SidharthK2 make sure the final wiki object stores ids in linkedWikis. cuz the backend expects to store id of wikis rather than the wiki name

image

src/components/CreateWiki/EditorModals/EditWikiDetailsModal/LinkedWikisInput.tsx
```diff
@@ -99,6 +96,13 @@ const LinkedWikisInput = ({ wiki }: { wiki: Wiki }) => {
     wiki.linkedWikis &&
     Object.values(wiki.linkedWikis).some((wikis) => wikis.length > 0)
 
+  function getWikiPreviewByTitle(
```
Adebesin-Cell, line 99: Kindly make this an arrow function.

Adebesin-Cell (approved):
Cool.

PR #1524, opened 2023-11-09
Feat add founders table

SidharthK2:
# Adds founders rank table

- Adds Founder table to Rank page
- Adds Founder table to homepage

image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1927

### DIFF
--- a/src/components/Landing/RankingList.tsx
+++ b/src/components/Landing/RankingList.tsx
@@ -12,12 +12,22 @@ import {
 } from '@chakra-ui/react'
 import { useTranslation } from 'react-i18next'
 import { BiImage } from 'react-icons/bi'
-import { RiCoinsFill, RiRobotFill, RiCoinFill } from 'react-icons/ri'
+import {
+  RiCoinsFill,
+  RiRobotFill,
+  RiCoinFill,
+  RiUserFill,
+} from 'react-icons/ri'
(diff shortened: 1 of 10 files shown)


Simplypheyie96 (commented):
It looks okay without the wikis, well can only know if the table looks good once we see the stacked founders.

What to work on though is this extra space on mobile view after scrolling across the table @SidharthK2

Simplypheyie96:
It looks okay with the stacked images. When it exceeds more than two founders we break it into an ellipse.

Simplypheyie96 (changes requested):
there is still so much space in the founder's tab which makes the rest of the table look squeezed together, let's reduce the space before the next tab, also the table are not aligned properly
Screenshot 2023-11-09 at 2 52 38 PM

Screenshot 2023-11-09 at 2 50 29 PM

SidharthK2:
> there is still so much space in the founder's tab which makes the rest of the table look squeezed together, let's reduce the space before the next tab, also the table are not aligned properly Screenshot 2023-11-09 at 2 52 38 PM
> 
> Screenshot 2023-11-09 at 2 50 29 PM

image
Longer name rows might look messy if I reduce the gap too much.

Simplypheyie96 (approved):

Softdev1 (approved):

PR #1526, opened 2023-11-14
Fix rank table UI

SidharthK2:
# Various UI tweaks on the rank table

- Fixes founder tab inconsistencies
- Reduce image and font sizes
- Adjust layout and add image border
- Fix SEO preview (WIP)

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/1976


Simplypheyie96:
i don't think we can close out this issue yet. there is still a lot to fix design-wise based on the feedback that would need implementation that i just started working on today.

Simplypheyie96 (changes requested):
- I think this has to be reviewed. There are spaces on the sides and we might need to move the bar a little closer to the navs on top so the elements don't all look isolated.
image

if you look at the design, there are no spaces on the sides and all elements don't look isolated from each other.
image

- also marketcap netchange on founders should be like this with the small triangles:
image

- descriptions on headers for each tab is not right. For example, the title of founders have been changed alonge with description, so do check each description for the sections.
image
@SidharthK2

Simplypheyie96 (changes requested):
@SidharthK2 the small icons are for founders tab only without the numbers. just the polygons.
image

this shouldn't be more than 2 to 3 lines
image

Simplypheyie96 (approved):

PR #1616, opened 2024-01-31
Table of Contents fix

SidharthK2:
# ToC fix
- Make ToC scroll independently from content

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2219


Simplypheyie96 (approved):
There is no long table of content on dev, so I will test on prod after merge.

Softdev1 (approved):

PR #1624, opened 2024-02-05
fix wiki image sizing in wiki

Tolu-Mals:
# Summary
Fixes wiki image size to be the same in viewing mode and editing mode

<details>
<summary> Before </summary>

Screenshot 2024-02-05 at 4 00 18 PM
</details>

<details>
<summary> After </summary>

Screenshot 2024-02-05 at 4 20 39 PM

</details>

## Linked issues
closes https://github.com/EveripediaNetwork/issues/issues/2242

### DIFF
--- a/src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
+++ b/src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
@@ -12,9 +12,7 @@ export const customImageRender = ({
   return (
     <Image
       quality={95}
-      loader={
-        typeof src === 'string' && src.startsWith('http') ? cfLoader : undefined
-      }
+      loader={cfLoader}
       src={src || ''}
       alt="wiki"
       width={900}
--- a/src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
+++ b/src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
@@ -1,5 +1,6 @@
-import { Image } from '@/components/Elements/Image/Image'
+import Image from 'next/image'
 import React from 'react'
+import { cfLoader } from '@/components/Elements/Image/Image'
 
 export const customImageRender = ({
   ...props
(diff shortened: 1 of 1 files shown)


src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
```diff
@@ -1,19 +1,27 @@
...
-      my={2}
-      objectFit="contain"
-      src={props.src || ''}
+      quality={95}
+      loader={cfLoader}
+      src={src || ''}
       alt="wiki"
+      width={900}
+      height={400}
+      style={{
```
Aliiiu, line 20: Why are you using the image prop and inline-style to set the image dimension

Aliiiu (changes requested):

src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
Tolu-Mals: Nextjs images require that you pass a specific width and height for their optimizations. However since I want the actual values to be 'auto', I specified that inline. If I specified only via inline styles, it'll throw an error, because width and height props are required, but If I specified 'auto' via props, it'll throw an error, cos it expects actual values.

Tolu-Mals: What I'm trying to do is let the image use the intrinsic width and height, you do that by setting the value of width and height to 'auto'. The width and height props don't support that.

Aliiiu (approved):

src/components/Wiki/WikiPage/CustomRenderers/customImageRender.tsx
Aliiiu: Yes. Check i think using the fill prop of the next image can help with that.

Softdev1 (approved):
I think we are good

PR #1631, opened 2024-02-08
Wiki Star Rating

SidharthK2:
# Adds wiki star rating feature

- Replaces thumbs up/down system with star rating for wikis

image

## WIP
- Integrate with API

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2254


Simplypheyie96 (changes requested):
- Implementation is different from design.
It is quite important for the stars under average rating to have the yellow stars (what typical rating stars look like), to differentiate it from the the stars the users will click to rate the wiki.
- The distance between the stars is too large(both for rating and showcase of average rating)
- the average rating stars are too tiny, while the "click to rate stars" are not big enough.
---
Implementation should look like this.
ps: also check for the no average rating state: https://www.figma.com/file/wosuJaHeV318k1ON82CZpo/%F0%9F%A7%A0-IQ.wiki?type=design&node-id=30394-37495&mode=design&t=WkwrMxs2cdxmErvG-11
image

Simplypheyie96 (changes requested):
This is not right still.
The stars under average ratings are the ones in yellow, The main stars where users will click will be the one in pink.
and there is still so much space between the smaller stars.
image

SidharthK2:
> @SidharthK2 i think we should have a few minutes of delay after users vote so that they see their vote before showing feedback. A simple graze over the feedback modal votes instantly without the user seeing what they did especially on mobile.

Not sure I understand, could you elaborate please?

Simplypheyie96:
> > @SidharthK2 i think we should have a few minutes of delay after users vote so that they see their vote before showing feedback. A simple graze over the feedback modal votes instantly without the user seeing what they did especially on mobile.
> 
> Not sure I understand, could you elaborate please?

Problem: On mobile, users can accidentally vote and see feedback message immediately, without realizing what they voted for while scrolling. 
solution: Wait a few seconds after someone votes before showing the feedback. This gives them time to see what they chose and confirm their vote.
this was why I had a submit button in the design previously, but if we are removing the submit button, let's make sure we give users a few mins to see the result of their vote before we switch to the feedback message.

SidharthK2:
@Simplypheyie96 How are we handling already voted user revisiting a wiki?

Simplypheyie96:
> @Simplypheyie96 How are we handling already voted user revisiting a wiki?

Since they already voted, they see just the current average votes the wiki has.

SidharthK2:
> > @Simplypheyie96 How are we handling already voted user revisiting a wiki?
> 
> Since they already voted, they see just the current average votes the wiki has.

So we are not allowing change of vote?

Simplypheyie96:
> > > @Simplypheyie96 How are we handling already voted user revisiting a wiki?
> > 
> > Since they already voted, they see just the current average votes the wiki has.
> 
> So we are not allowing change of vote?

I don't think that's necessary. Is it? 🫠
I don't think they should be voting multiple times.

Softdev1:
Le's consider alignment in implementation 
image

Simplypheyie96 (approved):

Softdev1 (approved):
Looks good now

Aliiiu (approved):

PR #1666, opened 2024-02-22
Event api integration

Aliiiu:
# Event API Integration

This PR works on the integration and testing of the event API

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2301


Simplypheyie96 (changes requested):
- event suggestion modal has wiki input details in it
image
---
- we need a clear all for the filter once they are selected and users wants to clear the selection. (clear all only appears on filter selection).
image
---
the media arrow is implemented for opposite functions on open/close
image

AshakaE (changes requested):
1. I believe we'll need a button that takes us to events page
2. interests 'Blockchain' and tag 'Blockchains' will miss matches

| | |
| --- | --- |
| image | A
image | B

3. Using only blockchain filter does not show tags on event card

|  |  |
| --- | --- |
| image | A
image | B

Aliiiu:
> 1. I believe we'll need a button that takes us to events page
> 2. interests 'Blockchain' and tag 'Blockchains' will miss matches
> 
> image	A
> image	B
> 3. Using only blockchain filter does not show tags on event card
> 
> image	A
> image	B

1. Will update the navbar to route to the page
2. Will resolve that thanks, nice catch
3. I will update the query for that

Simplypheyie96 (changes requested):
- Let's reduce the space between the sections indicated in red in the image below to maybe like 30px.
- let's also reduce the padding around the "interests" section.
The goal is to have at least one of the events show on a 14inches devices.
image
---
- Filter by next month doesn't seem to work.  it shows March instead of april.
image
---
- tags on events details page should open events like a search result but it navigates to the homepage and show no indication if it's working or not.
---
we should change the oreintation of this in md breakpoint because of all the space on the side. Or we put popular events together with nearby events beside it.
image

Simplypheyie96 (approved):

AshakaE (approved):

PR #1675, opened 2024-02-27
Fix media error

SidharthK2:
# Fixes media error
- Check for invalid youtube vids
- Prevent editor from exceeding media items limit of 25

## Notes or observations
currently checking for valid videos using a workaround to avoid using youtube API

```js
try {
    const response = await fetch(url, {
      method: 'GET',
      headers: {},
    })

    const text = await response.text()
    isUnavailable = text.includes('Video unavailable')

    res.status(200).json({ isUnavailable })
  } catch (error) {
    console.error('Error fetching video page:', error)
    res
      .status(500)
      .json({ error: 'Error fetching video page', isUnavailable: true })
  }
}
```

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2276

### DIFF
--- a/src/components/CreateWiki/EditorModals/MediaModal.tsx
+++ b/src/components/CreateWiki/EditorModals/MediaModal.tsx
@@ -35,7 +35,7 @@ import { WikiImageObjectProps } from '@/types/CreateWikiType'
 import { saveImage } from '@/utils/CreateWikiUtils/saveImage'
 import { useTranslation } from 'next-i18next'
 
-const MAX_MEDIA = 12
+const MAX_MEDIA = 25
 
 const MediaModal = ({
   onClose = () => {},
@@ -254,6 +254,7 @@ const MediaModal = ({
                 setImage={handleSetImage}
                 showFetchedImage={false}
                 modalUpload
+                media={wiki.media}
               />
             </Flex>
             {wiki.media !== undefined && wiki.media?.length > 0 && (
@@ -276,7 +277,7 @@ const MediaModal = ({
                         wiki.media?.length > MAX_MEDIA
                       }
                     >
-                      <Text fontSize="xs">Save {wiki.media.length}</Text>
+                      <Text fontSize="xs">Save</Text>
                     </Button>
                   </Tooltip>
                   <Button onClick={onClose} variant="outline" size="md">
--- a/src/components/CreateWiki/EditorModals/MediaModal.tsx
+++ b/src/components/CreateWiki/EditorModals/MediaModal.tsx
@@ -17,6 +17,7 @@ import {
   Stack,
   useToast,
   Select,
+  Tooltip,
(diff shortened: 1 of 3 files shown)


src/components/CreateWiki/EditorModals/MediaModal.tsx
```diff
@@ -253,14 +254,32 @@ const MediaModal = ({
...
+                      onClick={onClose}
+                      isDisabled={
+                        wiki.media === undefined ||
+                        wiki.media?.length <= 0 ||
+                        wiki.media?.length > MAX_MEDIA
+                      }
+                    >
+                      <Text fontSize="xs">Save</Text>
+                    </Button>
+                  </Tooltip>
```
Royal-lobster, line 282: @SidharthK2 we have a function for validating wiki contents and it automatically throws the error toasts if it finds any error, once check the code again if you can find it

Royal-lobster:
Also i believe we can solve this issue permanently by having common validation code for both frontend and backend inside `iq-utils` - https://github.com/EveripediaNetwork/iq-utils/ 

prolly a `validateWiki` function export that takes in wiki object and return [ValidatorCode](https://github.com/EveripediaNetwork/iq-utils/blob/531717d0d7043833a4766c2a11b55a63c44d846f/src/types/wiki.ts#L121) and may be detailed description cuz there are multiple ways a validator code can be invalid. 

By this we can just use the same function in frontend and backend and both results would match

src/components/CreateWiki/EditorModals/MediaModal.tsx
AshakaE: It's an extra step in validation, which imo is great. This will help editors to know they've exceeded media limit before it's time to publish
image

kesar:
> Also i believe we can solve this issue permanently by having common validation code for both frontend and backend inside `iq-utils` - https://github.com/EveripediaNetwork/iq-utils/
> 
> prolly a `validateWiki` function export that takes in wiki object and return [ValidatorCode](https://github.com/EveripediaNetwork/iq-utils/blob/531717d0d7043833a4766c2a11b55a63c44d846f/src/types/wiki.ts#L121) and may be detailed description cuz there are multiple ways a validator code can be invalid.
> 
> By this we can just use the same function in frontend and backend and both results would match

its a good suggestion but not sure what side effects will create. I think we can merge this and do that in the future 👍🏻

AshakaE (approved):

kesar:
looking forward to finally fix the infamous media error bug :)))

Royal-lobster (approved):

src/components/CreateWiki/EditorModals/MediaModal.tsx
Royal-lobster: Ahh i get it now, makes sense !

PR #1695, opened 2024-03-01
fix language switch bug

Tolu-Mals:
# Summary
Fixes bug where language switch shows wrong language

## How should this be tested?
1. type the language directly in the URL for example: iq.wiki/ko
2. Check to see the language switch selected language corresponds with the language being used

Closes https://github.com/EveripediaNetwork/issues/issues/2306

### DIFF
--- a/src/components/Layout/Navbar/Navbar.tsx
+++ b/src/components/Layout/Navbar/Navbar.tsx
@@ -18,9 +18,8 @@ import {
 import { CloseIcon, HamburgerIcon, ChevronDownIcon } from '@chakra-ui/icons'
 import { languageData } from '@/data/LanguageData'
 import dynamic from 'next/dynamic'
-import { useRouter } from 'next/router'
 import { useDispatch } from 'react-redux'
-import { setDrawerOpen } from '@/store/slices/app-slice'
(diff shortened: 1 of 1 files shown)


SidharthK2 (commented):
Unable to switch back to EN from KO by editing URL

Aliiiu:
> Unable to switch back to EN from KO by editing URL

I was able to by removing the "/ko"

Aliiiu:
image

@Tolu-Mals

SidharthK2:
image

Tolu-Mals:
> image @Tolu-Mals

Kindly share the steps to reproduce this

Aliiiu:
> > image @Tolu-Mals
> 
> Kindly share the steps to reproduce this

after adding "/ko" to the url. Then remove it to go back to en translation. The selector locale state does not change between those two state

Aliiiu:
> @Aliiiu @SidharthK2 can you check again, should be fixed now

The behavior is different, now i can switch from en to ko but i can't switch from ko to en by editing the url. When i try to switch from ko to en from the url it keeps persisting the ko lang

PR #1697, opened 2024-03-04
Closable switch network

Adebesin-Cell:
# Closable switch network

This PR ensures switch network doesn't open after being closed

Fixes https://github.com/EveripediaNetwork/issues/issues/2279

### DIFF
--- a/src/components/CreateWiki/CreateWikiTopBar/WikiPublish/WikiPublishButton.tsx
+++ b/src/components/CreateWiki/CreateWikiTopBar/WikiPublish/WikiPublishButton.tsx
@@ -42,7 +42,7 @@ const NetworkErrorNotification = dynamic(
 )
 
 export const WikiPublishButton = () => {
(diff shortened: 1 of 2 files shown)


src/components/CreateWiki/CreateWikiTopBar/WikiPublish/WikiPublishButton.tsx
```diff
@@ -331,6 +334,7 @@ export const WikiPublishButton = () => {
       <NetworkErrorNotification
         modalState={showNetworkModal}
```
Royal-lobster, line 335: here model state can be just `connectedChainId !== chainId && networkSwitchAttempted`. `networkSwitchAttempted` can start from false, and changes to true when the model is closed so this won't get triggered again, by this i think you don't need that useEffect at top

CleanShot 2024-03-05 at 09 44 29

Royal-lobster (approved):

PR #1713, opened 2024-03-07
fixed ui hydration error

invisiblemask:
# fixed ui hydration error

hydration error pops up when running this app on the local server, encountered error while trying to create a pull request

## How should this be tested?

1. start the app on the local server
2. refresh the page a couple times to ensure the error is properly tested

## Notes or observations

discovered .husky code differs across different machine


invisiblemask:
modified this pr, changing /common.sh to _/husky.sh"

Tolu-Mals:
The `pre-commit.sh` file is now back to the way it was initially, and it's not a good idea to suppress hydration warnings, in case someone wants to attempt to fix it. We can close this

PR #1722, opened 2024-03-07
Fixed husky code configuration

invisiblemask:
# Fixed husky code configuration

made a mistake while configuring .husky pre-commit, fix that by adding extra line of code

changed this
Screenshot (61)

to this
Screenshot (62)

## Linked issues

closes #00, closes #001, closes https://github.com/EveripediaNetwork/issues/issues/ISSUE_NUMBER_GOES_HERE


.husky/pre-commit.sh
```diff
@@ -0,0 +1,5 @@
+# .husky/pre-commit
+. "$(dirname -- "$0")/common.sh"
```
kesar, line 2: why do we had _/husky.sh and now common.sh and it was working for other env. ?

also if this works for everyone, there is other husky files in the repo that might follow same change, not only this one.

before merging lets ensure everything work for others as well

kesar:
you have already a PR opened, you can modify that one (or close it)

.husky/pre-commit.sh
invisiblemask: this is due to the error i encountered while creating a pr, since there are other husky files in the repo i will have to change it back to _/husky.sh fix the error and modify the previous pr

PR #1727, opened 2024-03-12
Fixes Magic Bundle

Adebesin-Cell:
# Fixes Magic Bundle

This PR investigates and removes extra bundling from magic that should stay in login.

Fixes https://github.com/EveripediaNetwork/issues/issues/2364

### DIFF
--- a/src/components/Elements/Avatar/DisplayAvatar.tsx
+++ b/src/components/Elements/Avatar/DisplayAvatar.tsx
@@ -5,9 +5,9 @@ import {
   Box,
   Center,
   ChakraProps,
-  CSSObject,
   HTMLChakraProps,
   Icon,
+  SystemStyleObject,
 } from '@chakra-ui/react'
 import { AvatarColorArray } from '@/data/AvatarData'
 import { RiUserLine } from 'react-icons/ri'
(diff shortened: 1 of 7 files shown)


src/components/Elements/Avatar/DisplayAvatar.tsx
```diff
@@ -48,6 +49,8 @@ const DisplayAvatar = ({
     }
   }, [address, avatarIPFS, setAccount])
 
+  console.log()
```
kesar, line 52:
```suggestion
```

src/pages/_document.tsx
```diff
@@ -26,8 +25,6 @@ class MyDocument extends Document<Props> {
             content="black-translucent"
           />
           <link rel="manifest" href="/manifest.json" />
-          <link rel="preconnect" href="https://i.ytimg.com" />
```
Tolu-Mals, line 29: this guy isn't related to magic, its supposed to increase performance on wiki pages by making the video thumbnails load faster. If you can include this only wiki pages, please do.

src/pages/_document.tsx
```diff
@@ -26,8 +25,6 @@ class MyDocument extends Document<Props> {
...
-          <link rel="preconnect" href="https://assets.auth.magic.link" />
```
Tolu-Mals, line 30: Also if you can include this guy only on the login page, it's a performance optimization.

src/pages/_document.tsx
Adebesin-Cell: got it!

src/pages/_document.tsx
Adebesin-Cell: Noted ser

Royal-lobster (approved):

Adebesin-Cell:
> @Adebesin-Cell what is pending on this pr provide enough reason

There's no pending tasks, it's done.

PR #1737, opened 2024-03-14
Fixes switch chain bug

Adebesin-Cell:
# Fixes switch chain bug

This PR upgrades wagmi, viem and magic labs latest versions, also maintains a mini code for magic connector pending the release of new magiclab connector.

Fixes https://github.com/EveripediaNetwork/issues/issues/2382
Fixes https://github.com/EveripediaNetwork/issues/issues/2349


package.json
```diff
@@ -37,7 +37,7 @@
     "@everipedia/iq-utils": "^0.2.3",
     "@everipedia/web3-signer": "^0.0.11",
     "@fortaine/fetch-event-source": "^3.0.6",
-    "@magiclabs/wagmi-connector": "^1.1.0",
+    "@magiclabs/wagmi-connector": "^1.1.5",
```
Royal-lobster, line 40: @Adebesin-Cell This should be a minor update acc to semantic versioning since here the pear dependencies change, you can change it in your pr in magic repo too to make it 1.2.0

Adebesin-Cell: 1.1.5 is the current value :))

Adebesin-Cell: 1.1.5 is the current value :))

src/config/wagmi.ts
```diff
@@ -1,61 +1,52 @@
...
-      customLogo: '/images/logos/braindao-logo.svg',
-      accentColor: '#ea3b87',
-      magicSdkConfiguration: {
-        network: {
-          rpcUrl: rpcs[config.alchemyChain],
-          chainId: Number(config.chainId),
+      relayUrl: 'wss://relay.walletconnect.org',
+    }),
+    dedicatedWalletConnector({
+      //@ts-ignore
```
Royal-lobster, line 35: Still showing type errors ?

Royal-lobster (approved):

src/config/wagmi.ts
Adebesin-Cell: Yep! some type mismatch and I didn't want to duplicate the chain array.

Royal-lobster: Ahh. tho try to add a proper fix for this, even if it works fine with suppressing the type error, probably there is a chance it would be a foot gun in some other cases

Adebesin-Cell: Fixed!

PR #1742, opened 2024-03-15
fixed multiple rating for a wiki post

invisiblemask:
# Fixed multiple rating for a wiki post

_On the rating widget, it was noticed that users can rate a wiki multiple times upon refresh of that page. I fixed the issue allowing a user on a single to just rate the wiki once and shouldn't be able to rate the wiki again ._

## How should this be tested?

1. Go to the IQ.wiki page, click on any wiki article
2. Rate the article
3. Refresh the page or go back to previous page and select the wiki

closes https://github.com/EveripediaNetwork/issues/issues/2397


Tolu-Mals (approved):

Tolu-Mals:
@invisiblemask  you can merge this

PR #1750, opened 2024-03-19
Fix performance issue

invisiblemask:
# Fix performance issue

_The homepage loads duplicate images on page load and some images have large sizes, these issues cause performance problems on the homepage, the images were optimized by adding a next image tag to images that don't have one and handling dynamic rendering to components where those images are located_

## How should this be tested?

1. When the homepage loads, open the Chrome debugger console
2. Click on the network tab and observe how the image loads

closes https://github.com/EveripediaNetwork/issues/issues/2402


Tolu-Mals:
Nice work so far. One more thing we can do to improve performance

On the home page, the Trending Card Component (see the two cards in the screenshot), it's using our custom Image component that has the quality value for image optimization hard coded to 95%. 

Screenshot 2024-03-19 at 12 21 17 PM

For places like in the screenshot above, where the image is small, we should be able to bump the image quality down a lot (maybe down to 30%, you can play with the values to see what looks presentable). This should increase performance significantly without a noticeable difference in visuals.

I tried bumping down the quality for one of the images, the size goes from 175Kb to 19.2kb.

Screenshot 2024-03-19 at 12 44 08 PM

So instead of the quality value in our Custom Image component being hardcoded to 95%, make it assignable via prop and bump it down on smaller images.

invisiblemask:
> Nice work so far. One more thing we can do to improve performance
> 
> On the home page, the Trending Card Component (see the two cards in the screenshot), it's using our custom Image component that has the quality value for image optimization hard coded to 95%.
> 
> Screenshot 2024-03-19 at 12 21 17 PM
> For places like in the screenshot above, where the image is small, we should be able to bump the image quality down a lot (maybe down to 30%, you can play with the values to see what looks presentable). This should increase performance significantly without a noticeable difference in visuals.
> 
> I tried bumping down the quality for one of the images, the size goes from 175Kb to 19.2kb.
> 
> Screenshot 2024-03-19 at 12 44 08 PM
> So instead of the quality value in our Custom Image component being hardcoded to 95%, make it assignable via prop with a default value of 95%, and bump it down on smaller images

okay cool, I will work on it now!!

src/pages/index.tsx
```diff
@@ -13,14 +13,24 @@ import { Wiki } from '@everipedia/iq-utils'
...
+)
+const LeaderBoard = dynamic(() => import('@/components/Landing/Leaderboard'), {
+  ssr: false,
+})
 import { editorApi, getLeaderboard, LeaderBoardType } from '@/services/editor'
 import { sortLeaderboards } from '@/utils/DataTransform/leaderboard.utils'
 import { RankCardType } from '@/types/RankDataTypes'
-const RankingList = dynamic(() => import('@/components/Landing/RankingList'))
+const RankingList = dynamic(() => import('@/components/Landing/RankingList'), {
+  ssr: false,
```
kesar, line 32: if we change SSR of these components wouldnt that make homepage to load slower? 

have you benchmarked this somehow?

Softdev1:
@invisiblemask what is remaining on this issue

Tolu-Mals:
@invisiblemask confirm from the docs whether setting `ssr: true` isn't going to affect performance negatively. Also, I don't think setting ssr to true to would cause duplication of images, confirm that.

Tolu-Mals (approved):

PR #1764, opened 2024-03-27
Reverts wagmi optimization for making publishing work

Royal-lobster:
# Changes
- wrapps all pages in wagmi provider which should fix the connector not found issue


Adebesin-Cell:
There's a weird issue with wrapping the entire app with wagmi, it breaks the homepage :)) without a specific error. I did comment out all components during debug and it failed, only way it worked in homepage is after a sign in.

Royal-lobster:
Do we still need this PR ?

Adebesin-Cell:
> Do we still need this PR ?

i doubt we can close it.

PR #1781, opened 2024-04-02
Wrong updated time

Aliiiu:
# Wrong Updated Time

This PR works on updating the logic for updated time of featured wiki elements

## How should this be tested?

1. Update a featured wiki
2. Navigate to the home page to check the updated time

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2468


Tolu-Mals:
I think you might have misunderstood this issue. The issue was that it was showing the wrong date on the featured wikis, but this pr updates the translations, which I think were already fine.

I think instead you try to find out if it's showing the wrong dates on the home page, and if yes, fix it.

Aliiiu:
> I think you might have misunderstood this issue. The issue was that it was showing the wrong date on the featured wikis, but this pr updates the translations, which I think were already fine.
> 
> I think instead you should try to find out if it's showing the wrong dates on the home page, and if yes, fix it.

The issue originated from trying to translate the distanceFromNow. it assumed the unit is in days for all featured wikis. So if I make an update to a featured wiki it will return the right number but wrong unit. So I think the observation came from a newly edited wiki with the wrong distanceFromNow

Tolu-Mals:
> > I think you might have misunderstood this issue. The issue was that it was showing the wrong date on the featured wikis, but this pr updates the translations, which I think were already fine.
> > I think instead you should try to find out if it's showing the wrong dates on the home page, and if yes, fix it.
> 
> The issue originated from trying to translate the distanceFromNow. it assumed the unit is in days for all featured wikis. So if I make an update to a featured wiki it will return the right number but wrong unit. So I think the observation came from a newly edited wiki with the wrong distanceFromNow

ohh, okay. But I assume you wouldn't need to change translations to fix that. I'll test the preview link now

src/components/Landing/FeaturedWikiCard.tsx
```diff
@@ -129,7 +143,7 @@ export const FeaturedWikiCard = ({ wiki }: { wiki: Wiki }) => {
                 fontSize="sm"
                 textAlign="right"
               >
-                {t('LastEdited')} {getLatestEdited()} {t('days')}
```
Tolu-Mals, line 132: The issue here is it was hard coded to check for days. all you need to do is use `useTranslatedTimestamps` and that would fix the problem. No need for changing the logic for useTranslatedTimestamps and the translations, unless it doesn't work fine.

Tolu-Mals: Check the activity cards, those display the correct values

Aliiiu:
> > > I think you might have misunderstood this issue. The issue was that it was showing the wrong date on the featured wikis, but this pr updates the translations, which I think were already fine.
> > > I think instead you should try to find out if it's showing the wrong dates on the home page, and if yes, fix it.
> > 
> > 
> > The issue originated from trying to translate the distanceFromNow. it assumed the unit is in days for all featured wikis. So if I make an update to a featured wiki it will return the right number but wrong unit. So I think the observation came from a newly edited wiki with the wrong distanceFromNow
> 
> ohh, okay. But I assume you wouldn't need to change translations to fix that. I'll test the preview link now

The useTranslatedTimeStamps used the regex that does not work for all values return from the date-fns library. if I have "14 hours" it returns ["1", "4", "hours"] but if we have "1 hour" it returns ["1", "hour"].

The translation of the value return for other locales aside "en" is not similar to what we have for "en"

Edited 1 hour ago is translated using the number first and the translation is not similar to what we have for "en". So I broke the translation for each relativeDistance unit to follow the same words.

Tolu-Mals:
> > > > I think you might have misunderstood this issue. The issue was that it was showing the wrong date on the featured wikis, but this pr updates the translations, which I think were already fine.
> > > > I think instead you should try to find out if it's showing the wrong dates on the home page, and if yes, fix it.
> > > 
> > > 
> > > The issue originated from trying to translate the distanceFromNow. it assumed the unit is in days for all featured wikis. So if I make an update to a featured wiki it will return the right number but wrong unit. So I think the observation came from a newly edited wiki with the wrong distanceFromNow
> > 
> > 
> > ohh, okay. But I assume you wouldn't need to change translations to fix that. I'll test the preview link now
> 
> The useTranslatedTimeStamps used the regex that does not work for all values return from the date-fns library. if I have "14 hours" it returns ["1", "4", "hours"] but if we have "1 hour" it returns ["1", "hour"].
> 
> The translation of the value return for other locales aside "en" is not similar to what we have for "en"
> 
> Edited 1 hour ago is translated using the number first and the translation is not similar to what we have for "en". So I broke the translation for each relativeDistance unit to follow the same words.

oh, makes sense

Tolu-Mals:
I feel like we could even write unit tests for the useTranslatedTimestamps hook, maybe another pr. If I did that I could have caught the issue with the regex

Tolu-Mals:
One small fix, translations are missing 'ago'. eg. It's showing 'Last edited 4 mins' instead of 'Last edited 4 mins ago'

Aliiiu:
> One small fix, translations are missing 'ago'. eg. It's showing 'Last edited 4 mins' instead of 'Last edited 4 mins ago'

I don't think it was part of the previous content for featured wiki. They don't have "ago"

image

Tolu-Mals (approved):
Nice work ✨

Softdev1 (approved):

PR #1799, opened 2024-04-15
Add price to stats widget

Tolu-Mals:
# Summary

Adds price to the stats widget on wiki pages

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2495


Simplypheyie96 (approved):

Aliiiu:
Hi, Can we also maintain the same number of decimal place for the price stats like we have in the homepage?

image

Aliiiu (approved):

Tolu-Mals:
> Hi, Can we also maintain the same number of decimal place for the price stats like we have in the homepage?
> 
> image

ah, I missed this. will prolly include this in some other pr

PR #1830, opened 2024-05-14
Fixes hydration error on homepage

Royal-lobster:
Fixes https://github.com/EveripediaNetwork/issues/issues/2629

# Screenshots

https://github.com/EveripediaNetwork/ep-ui/assets/52039218/9ddea964-2fc7-4dd0-a187-46ad4d9f9926

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -13,6 +13,9 @@ import { Wiki } from '@everipedia/iq-utils'
 import TrendingWikis from '@/components/Landing/TrendingWikis'
 const CategoriesList = dynamic(
   () => import('@/components/Landing/CategoriesList'),
+  {
+    ssr: false,
+  },
 )
 import { getTags, tagsApi } from '@/services/tags'
 const DiscoverMore = dynamic(() => import('@/components/Landing/DiscoverMore'))


src/pages/index.tsx
```diff
@@ -13,6 +13,9 @@ import { Wiki } from '@everipedia/iq-utils'
 import TrendingWikis from '@/components/Landing/TrendingWikis'
 const CategoriesList = dynamic(
   () => import('@/components/Landing/CategoriesList'),
+  {
+    ssr: false,
```
kesar, line 17: does this increase latency / extra api calls that slowdown homepage?

Royal-lobster: Oh, i think it won't cause extra api calls cuz the data is being passed down as props, but the drawback i see is this part of the section won't render on server - so bots can't see the html of categories. tho not google cuz google bot do render some javascript so in theory it shouldn't effect SEO.

kesar: not worried about SEO since its on the sitemaps, as long as it doesnt do extra api calls to load those categories we are good 👍🏻

kesar (approved):

PR #1845, opened 2024-05-21
Blog improvements

kesar:
fixes https://github.com/EveripediaNetwork/issues/issues/2595

### DIFF
--- a/src/pages/blog/[digest].tsx
+++ b/src/pages/blog/[digest].tsx
@@ -11,7 +11,7 @@ import {
   Link,
 } from '@chakra-ui/react'
 import { store } from '@/store/store'
-import { formatBlog, getBlogsFromAllAccounts } from '@/utils/blog.utils'
+import { formatBlog, getBlogsFromAccounts } from '@/utils/blog.utils'
 import ReactMarkdown from 'react-markdown'
 import { components, uriTransformer } from '@/components/Blog/BlogMdComponents'
 import remarkParse from 'remark-parse'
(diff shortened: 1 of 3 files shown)


src/pages/blog/[digest].tsx
```diff
@@ -169,7 +169,11 @@ export const BlogPostPage = ({
 
 export const getStaticProps: GetStaticProps = async (context) => {
   const digest: string = context.params?.digest as string
-  const result = await store.dispatch(getEntry.initiate(digest))
+  const [result, translations, blogEntries] = await Promise.all([
```
Comment (kesar) [inline src/pages/blog/[digest].tsx:172]: lets remember to try to put together blocking calls together so we improve time of generation

src/pages/blog/index.tsx
```diff
@@ -46,22 +46,18 @@ export const Blog = ({ blogEntries }: { blogEntries: BlogType[] }) => {
 }
 
 export const getStaticProps: GetStaticProps = async ({ locale }) => {
-  const blogEntries = await getBlogsFromAllAccounts()
-  await Promise.all(store.dispatch(ArweaveApi.util.getRunningQueriesThunk()))
-
-  blogEntries?.sort((a, b) => {
```
kesar, line 52: this sort looks better inside of the getBlogsFromAccount so the page compontent is smaller and more readable

src/utils/blog.utils.ts
```diff
@@ -55,22 +55,29 @@ export const formatBlog = (blog: Blog, hasBody?: boolean) => {
...
-  let blogs: FormatedBlogType[] = []
+export const getBlogsFromAccounts = async () => {
   const accounts = [config?.blogAccount2, config?.blogAccount3]
 
-  // eslint-disable-next-line no-plusplus
-  for (let i = 0; i < accounts.length; i++) {
+  const blogPromises = accounts.map(async (account) => {
     // eslint-disable-next-line no-await-in-loop
-    let { data: entries } = await store.dispatch(getBlogs.initiate(accounts[i]))
+    const { data: entries } = await store.dispatch(getBlogs.initiate(account))
```
kesar, line 63: avoid loops w awaits instead create promises and do a promise all in the end, so everything runs in parallel . in this case its a 3x improvement

src/pages/blog/[digest].tsx
Comment (Tolu-Mals) [reply src/pages/blog/[digest].tsx:172]: Alright sir

src/pages/blog/index.tsx
Tolu-Mals: That's true

Tolu-Mals (approved):

PR #1852, opened 2024-05-22
User Profile Editing Bug

Aliiiu:
# User Profile Editing Bug

This PR works on resolving issues with users not being able to edit their details on the settings page

## How should this be tested?

1. Got to the settings page
2. Change any of the feeds

## Notes or observations

https://github.com/EveripediaNetwork/ep-ui/assets/67792966/2eb45ee9-e4ce-4fc3-9c62-59459b641647

## Linked issues
closes https://github.com/EveripediaNetwork/issues/issues/2660


Damola18:
Tested, this looks good.

Damola18 (approved):

Softdev1 (changes requested):
Not yet fixed. Username is not updated and still showed "profile updated success toss"

Aliiiu:
> Not yet fixed. Username is not updated and still showed "profile updated success toss"

Tested it, it is working

https://github.com/EveripediaNetwork/ep-ui/assets/67792966/097e18ea-8e65-4ffb-a6fc-38a1c7dd3072

Softdev1:
Screenshot 2024-05-26 at 16 00 36
this is the error i got

Aliiiu:
> Screenshot 2024-05-26 at 16 00 36 this is the error i got

Now we can see the error, which wasn't the case before. The issue is from the backend and we need to resolve it. 

@AshakaE it is still not working

Softdev1 (approved):

Softdev1:
Test-case for User Profile Editing Bug 

- User can now  edit profile details including Name, email, and other information
- Upon clicking the save button and refreshing the page, the entered information is updated.
- issues such as duplicate info have been fixed.

All test cases passed now welldone @Aliiiu  👍

src/components/Settings/ProfileSettings/index.tsx
```diff
@@ -72,6 +73,31 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
     }
   }, [settingsData, userENSAddr])
 
+  useEffect(() => {
```
Adebesin-Cell, line 76: I think we don't need an effect to handle errors, it could be handled in the handleSubmit handler :))

src/components/Settings/ProfileSettings/index.tsx
```diff
@@ -72,6 +73,31 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
...
+      toast({
+        title: toastTitle,
+        description: toastMessage,
+        status: toastType,
+        duration: 5000,
+        isClosable: true,
+      })
+    }
+    if (postProfileSuccess) {
+      const { toastTitle, toastMessage, toastType } = PostUserMessage(undefined)
```
Adebesin-Cell, line 90: Instead of passing undefined here, we can just call PostUserMessage and pass postProfileError, since it might be undefined hence it'll return the success message.

package.json
```diff
@@ -80,7 +80,7 @@
     "framer-motion": "^6.5.1",
     "fuse.js": "^6.6.2",
     "graphql": "^15.8.0",
-    "graphql-request": "^3.7.0",
+    "graphql-request": "4.3.0",
```
Adebesin-Cell, line 83: Did we need a version bump?

Aliiiu: yes, was unable to handle custom errors with the current version.

Adebesin-Cell: Ha, okay 👍

kesar, line 83: take in consideration that upgrade to a major release requires a full testing of the site. Have u consider what are the differences between 3.7 and 4.3 version of this library and retro compatibilities?

cc: @Softdev1 if you are testing this branch, you need to ensure everything works as expected not only the save because this is changing a library that its used across every single request.

Aliiiu: Error handling on the previous version is not dynamic, we can't extract error messages. We can't also parse the error string. Using the new version, we can create a customer error to extract error message.

Aliiiu: > Error handling on the previous version is not dynamic, we can't extract error messages. We can't also parse the error string. Using the new version, we can create a customer error to extract error message.

it was a bug that was fixed on that version

kesar: also that version is 2y old. i wonder if we should update to 7 if we aim to refactor that part, but if so, most likely the right way of doing this is: 

- fix current issue w current grapqhl-request ( i doubt we need to update version to catch an error, then the whole site wouldnt be working) .
- create a follow up task for refactoring and use the latest graphql-request version

Aliiiu: Updating to the latest version breaks a lot of things in the codebase. 4.3.0 was the one that resolved the issue. I will revert and create follow-up PR.

kesar: okok nvm. lets then keep 4.3.0 but lets ensure we test everything else. 

I dont want to fix the update profile bug that its barely used by anyone and get a bug somewhere else where its a critical component.

src/components/Settings/ProfileSettings/index.tsx
Aliiiu: The reason why I used a useEffect is because all request returns status 200 and the error won't be watched if I wrap the mutation in a try and catch.

Adebesin-Cell: Ha, weird, but it's prolly meant to return error object perhaps, or @AshakaE ?

Aliiiu: He said it's a graphql issue, all status is 200 for successful or failed request but the response is different for both

Adebesin-Cell: FlaggingSystemModal we can use the same approach here then I believe.

Aliiiu: @Adebesin-Cell resolved can you review again

src/components/Settings/ProfileSettings/index.tsx
```diff
@@ -140,20 +134,31 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
...
+      const { toastTitle, toastMessage, toastType } = PostUserMessage(errMsg)
+      toast({
+        title: toastTitle,
+        description: toastMessage,
+        status: toastType,
+        duration: 5000,
+        isClosable: true,
+      })
+    } else {
+      console.log(response.error)
```
Adebesin-Cell, line 149:
```suggestion
      console.log(response.error)
```

src/components/Settings/ProfileSettings/index.tsx
```diff
@@ -140,20 +134,31 @@ const ProfileSettings = ({ settingsData }: ProfileSettingsProps) => {
...
-    toast({
-      title: toastTitle,
-      description: toastMessage,
-      status: toastType,
-      duration: 5000,
-      isClosable: true,
-    })
-    setIsLoading(false)
+    const response = await postUserProfile({ profileInfo: data })
+    let errMsg
```
Adebesin-Cell, line 138: prolly make it string | undefined as type not any :))

src/components/Settings/ProfileSettings/index.tsx
Adebesin-Cell: for sure!

PR #1862, opened 2024-05-28
remove console log from revalidate api

invisiblemask:
# Remove log output from console

_Error from revalidate API is been in the console, this is because the error was being logged on the console, to fix, console.log from the API revalidate code was removed_

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2706


src/pages/api/revalidate.ts
```diff
@@ -19,7 +19,6 @@ export default async function handler(
     await res.revalidate(path)
     return res.json({ revalidated: true, path })
   } catch (err) {
-    console.error(JSON.stringify(err))
     return res.status(500).json({ path: path, error: JSON.stringify(err) })
```
kesar, line 22: i think its also because of sending error 500. probably we can send other type of error doesnt get into vercel error console. we can also keep the console.log but move it to console.warning()

invisiblemask: Okay sir, thank you 🙏

PR #1883, opened 2024-06-13
Refactor whitelist hook

SidharthK2:
## Refactor whitelist hook
- Run only if address if defined

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2771

### DIFF
--- a/src/hooks/useWhiteListValidator.ts
+++ b/src/hooks/useWhiteListValidator.ts
@@ -25,14 +25,6 @@ const useIsWhitelistedEditor = (): UseIsWhitelistedEditorRes => {
     },
   })
 
-  console.log(
-    'isWhitelisted:',
-    isWhitelisted,
-    'error:',
-    error,
-    'isLoading:',
-    isLoading,
-  )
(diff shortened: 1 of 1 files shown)


src/hooks/useWhiteListValidator.ts
```diff
@@ -1,26 +1,48 @@
...
     args: [address as `0x${string}`],
+    query: {
+      enabled: !!address,
+    },
   })
-  useEffect(() => {
-    if (isError) {
-      console.error('Error fetching whitelist data')
+
+  const handleContractRead = useCallback(() => {
```
Adebesin-Cell, line 28: maybe name should getContractReadError or logisWhitestedError

Adebesin-Cell (approved):

Softdev1:
Test-case for this PR  UI check. 

- When a user is not log-in or logout into the system, it should show "Suggest wiki."
image

- Once the user is log-in, it should show "create wiki"
image

Softdev1 (approved):
This is good to me

PR #1902, opened 2024-06-20
Fix inavlid url error

Damola18:
# Fix Invalid Url error and incomplete urls on explorers List

This PR fixes the solscan url and other urls that are not been added to the explorers list. It also updates tronscan regex from tronscan.com  which throws invalid url format 

# Before
image

image

# After

https://github.com/EveripediaNetwork/ep-ui/assets/50779080/38688684-4f73-4bb1-8163-1e3f94bafb8e

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2809

### DIFF
--- a/src/data/WikiLinks.ts
+++ b/src/data/WikiLinks.ts
@@ -233,6 +233,7 @@ export const LINK_OPTIONS = [
     type: LinkType.EXPLORER,
     label: 'Tronscan',
     icon: FaFileContract,
-    tests: [/https:\/\/(www.)?tronscan.com\/\w+/],
+    tests: [/https:\/\/(www\.)?tronscan\.org\/#\/token20\/\w+/],
+
   },
 ]
--- a/src/data/WikiLinks.ts
+++ b/src/data/WikiLinks.ts
@@ -234,6 +234,5 @@ export const LINK_OPTIONS = [
     label: 'Tronscan',
     icon: FaFileContract,
     tests: [/https:\/\/(www\.)?tronscan\.org\/#\/token20\/\w+/],
-
   },
 ]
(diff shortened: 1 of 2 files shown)


src/data/WikiLinks.ts
```diff
@@ -233,6 +233,6 @@ export const LINK_OPTIONS = [
     type: LinkType.EXPLORER,
     label: 'Tronscan',
     icon: FaFileContract,
-    tests: [/https:\/\/(www\.)?tronscan\.org\/#\/token20\/\w+/],
+    tests: [/https:\/\/(www\.)?tronscan\.(io|org)\/#\/\w+/],
```
Tolu-Mals, line 236: I'm not sure this is correct, you're checking for `/#/` in the links? Can you share sample links

Damola18: > I'm not sure this is correct, you're checking for `/#/` in the links? Can you share sample links

https://tronscan.org/#/token20/TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t
https://tronscan.io/#/transaction/3186f41dcbc51538082afe2ffd02f5f8e5b79d2d68a3d7e9fde74ffa4893ec9a

Tolu-Mals: thx. this might be good enough, but also consider checking for (token20|transaction), if those are the only two options.

Tolu-Mals (approved):

src/data/WikiLinks.ts
Damola18: Oh, forgot to respond to this , i noticed there are more couple addresses without /token20 and /transaction, so i think adding this might break the regex later in an event where there are still others that were not added.

PR #1921, opened 2024-07-05
Properly formatted the gas limit error and also resolved overflow on …

Yomzy2223:
# PR Overflow on error

A fix to ensure error message do not overflow

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2857


src/components/CreateWiki/EditorModals/WikiProcessModal.tsx
```diff
@@ -78,7 +81,7 @@ const WikiProcessModal = (props: WikiProcessType) => {
                     state={state}
                     labelOrientation="vertical"
                     colorScheme="pink"
-                    activeStep={activeStep}
```
kesar, line 81: are u sure about this change? not sure how it related to the issue

src/components/CreateWiki/EditorModals/WikiProcessModal.tsx
```diff
@@ -40,6 +40,9 @@ const WikiProcessModal = (props: WikiProcessType) => {
   ]
   if (!isOpen) return null
 
+  const messageInMsg = /"message"\:\s+"([^"]+)"/
+  const message = msg.match(messageInMsg)
```
kesar, line 44: now we have msg message and messageInMsg in the same component. This looks very confusing.

create a function outside of the component that receives a msg and return the proper message. Also ensure this should be done at this level and not in the previous level because we are already passing msg in props which leads me to think that perhaps the refining of msg needs to be done in the previous step.

kesar (changes requested):

src/components/CreateWiki/EditorModals/WikiProcessModal.tsx
Yomzy2223: The messageInMsg helps match whether or not there is a message in the msg coming from the props, in the gas limit error message instance for example. I'll create a function outside the component as you said, will receive the message and return a proper message as you said.

Screenshot 2024-07-04 at 16 35 44

src/components/CreateWiki/EditorModals/WikiProcessModal.tsx
Yomzy2223: Oh, made this change just for testing. Corrected now.

kesar (approved):

Adebesin-Cell (approved):

src/hooks/useGetSignedHash.ts
```diff
@@ -23,11 +23,17 @@ import { usePostHog } from 'posthog-js/react'
...
     if (firstError.extensions?.exception?.reason) {
-      return firstError.extensions.exception.reason
+      msg = firstError.extensions.exception.reason
     } else if (firstError.message) {
-      return firstError.message
+      msg = firstError.message
     }
+
+    const hasInnerMsg = msg.match(/"message"\:\s+"([^"]+)"/)
+    const properMsg = hasInnerMsg ? hasInnerMsg[1] : msg
```
Adebesin-Cell, line 34: perhaps we rename variables, message instead of message, then proper message can be extractedMessage?

Yomzy2223: True. Will look into that now. Thanks.

Softdev1:
When writing a pull request (PR), it is important to provide a proper title, description, and a guide on how to test the PR.@Yomzy2223 
image

Softdev1 (approved):

PR #1930, opened 2024-07-10
Add Rank Table link to Tags

SidharthK2:
## Adds rank table link to tags page

- Adds "view more" button on tags page that links to corresponding rank category

WIP
- Better solution for linking rank categories and tags

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2416


Simplypheyie96 (changes requested):
- It works great for founders and cryptocurrencies, but not for stablecoins. We have stablecoins on the rank table, so the tag should redirect there.

- The button design is also not right. We don't use rounded corner buttons on IQ.wiki.
The button according to design:
image

Simplypheyie96 (approved):

Softdev1 (approved):
works perfectly fine for me

PR #1942, opened 2024-07-16
fixed bug with wiki report submission

Yomzy2223:
# Wiki report submission bug

_Fixed bug with wiki report submission_

## How should this be tested?

1. Click on "Report to us" button from wiki details
2. Write in some texts then click "Confirm"

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2911


Adebesin-Cell (commented):
Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.

Yomzy2223:
> Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.

Resulted in parameter overload, as the backend only takes report and wikiId

Adebesin-Cell:
> > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> 
> Resulted in parameter overload, as the backend only takes report and wikiId

Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.

AshakaE:
> > > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> > 
> > 
> > Resulted in parameter overload, as the backend only takes report and wikiId
> 
> Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.

It's not needed, if you want to do a more robust feedback, you can use contentFeedback api

```gql
contentFeedback(
contentId: String
userId: String
site: ContentFeedbackSite!
message: String
rating: Int
input: String
output: String
reportType: String
): Boolean!
```

Adebesin-Cell:
> > > > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> > > 
> > > 
> > > Resulted in parameter overload, as the backend only takes report and wikiId
> > 
> > 
> > Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.
> 
> It's not needed, if you want to do a more robust feedback, you can use contentFeedback api
> 
> ```graphql
> contentFeedback(
> contentId: String
> userId: String
> site: ContentFeedbackSite!
> message: String
> rating: Int
> input: String
> output: String
> reportType: String
> ): Boolean!
> ```

Perhaps we use this instead of having two separate endpoints that does same thing but one is a smaller version of the other.

Yomzy2223:
> > > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> > 
> > 
> > Resulted in parameter overload, as the backend only takes report and wikiId
> 
> Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.

I have confirmed with @AshakaE, userId is not needed.

Adebesin-Cell:
> > > > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> > > 
> > > 
> > > Resulted in parameter overload, as the backend only takes report and wikiId
> > 
> > 
> > Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.
> 
> I have confirmed with @AshakaE, userId is not needed.

Yep! Got it. Let’s use content feedback

Yomzy2223:
> > > > > Why we remove userId? I think it's a necessary data to keep, we could make it optional if it's not meant to be recorded.
> > > > 
> > > > 
> > > > Resulted in parameter overload, as the backend only takes report and wikiId
> > > 
> > > 
> > > Perhaps confirm with @AshakaE about the user id, cos it seems to be passed as a required param.
> > 
> > 
> > I have confirmed with @AshakaE, userId is not needed.
> 
> Yep! Got it. Let’s use content feedback

Alright

public/locales/en/home.json
invisiblemask, line 1: This is not related to this PR

invisiblemask (commented):
There are so many unrelated changes on this PR, we can limit this PR to it's description

public/locales/en/home.json
Yomzy2223: I pulled & merged changes from main branch. That's why. I'll just create a fresh pr now, for just the fix alone, so it's easier to review.

invisiblemask: Oh okay,or you could revert other commits and leave yours?

PR #1956, opened 2024-07-22
Wiki crashing

Yomzy2223:
# Wiki crashing

## How should this be tested?

1. Try switching language to KO then back to EN

## Notes or observations

The error has to do with how next is handling navigation for the index page, it does not fetch the data from the server just returns undefined like the getServerSide code is absent.  That is the reason why the default state for the api calls in the server does not work. Setting default state for the page props will keep the page in a bad UI state since the data is stuck at the empty state and no api calls to update the state. The error disappears after some time when you click the logo, but it happens again when you click it. So the solution is to change how we are handling the navigation on clicking the logo

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2928

### DIFF
--- a/src/components/Landing/RankingList.tsx
+++ b/src/components/Landing/RankingList.tsx
@@ -49,13 +49,12 @@ type RankingListProps = {
 }
 
 const RankingList = ({ rankings, listingLimit }: RankingListProps) => {
-  const {
-    TokensListing,
-    aiTokensListing,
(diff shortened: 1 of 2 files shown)


SidharthK2 (commented):
Getting this
image

SidharthK2 (approved):

Softdev1 (approved):

PR #1961, opened 2024-07-23
fixed twitter widget issue

Yomzy2223:
# Twitter widget issue fix

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2939

### DIFF
--- a/src/components/Wiki/WikiPage/InsightComponents/TwitterTimeline.tsx
+++ b/src/components/Wiki/WikiPage/InsightComponents/TwitterTimeline.tsx
@@ -11,7 +11,7 @@ const TwitterTimeline = ({ url }: { url: string }) => {
   const [snapOpen, setSnapOpen] = useState(false)
   const [isLoaded, setIsLoaded] = useState(false)
 
-  const twitterLink = url?.replace(/x.com/gi, 'twitter.com')
+  const twitterLink = url.replace(/x.com/gi, 'twitter.com')
 
   useEffect(() => {
     const observer = new IntersectionObserver(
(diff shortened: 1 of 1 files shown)


src/components/Wiki/WikiPage/InsightComponents/TwitterTimeline.tsx
```diff
@@ -71,7 +71,7 @@ const TwitterTimeline = ({ url }: { url: string }) => {
                 noScrollbar
                 tweetLimit={4}
                 borderColor={colorMode === 'dark' ? '#4a5568' : '#ddd'}
-                url={url}
+                url={url.replace('x.com', 'twitter.com')}
```
Adebesin-Cell, line 74: See if you can parse it on input (on link addition) to twitter.com if that's what works, and move this processing a bit up, perhaps as a variable after isLoaded state.

Yomzy2223: True, will move the processing up. However, I don't think parsing it on input will be a nice idea. Seeing from the user's perspective, I'll feel confused if I enter my link as "https://x.com..." and it changes to "https://twitter.com..."

What do you think?

Adebesin-Cell: Hm, make sense. I was thinking since both urls are same but one seems to render the widget properly, we could just stick to one link instead of having two links of same form.

src/components/Wiki/WikiPage/InsightComponents/TwitterTimeline.tsx
```diff
@@ -11,6 +11,8 @@ const TwitterTimeline = ({ url }: { url: string }) => {
   const [snapOpen, setSnapOpen] = useState(false)
   const [isLoaded, setIsLoaded] = useState(false)
 
+  const twitterLink = url?.replace(/x.com/gi, 'twitter.com')
```
Adebesin-Cell, line 14: looks good, why optional chain? string is always going to be present.

Adebesin-Cell (approved):

Softdev1 (approved):

PR #1977, opened 2024-07-24
Resolving EditWiki Validation Error

Aliiiu:
# Resolving EditWiki Validation Error

This PR works on resolving the validation error when a user tries to add event details for a wiki

## How should this be tested?

1. Navigate to the create-wiki page
2. Click on Edit wiki
3. Try to add an event wiki details 

## Observation

Before
image

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2946


Softdev1 (approved):
image

src/components/CreateWiki/EditorModals/EditWikiDetailsModal/EventsInput/index.tsx
```diff
@@ -244,12 +244,22 @@ const EventsInput = ({ wiki }: { wiki: Wiki }) => {
                 type="text"
                 placeholder={t('title')}
               />
-              <ComboBoxPopup
```
Yomzy2223, line 247: Any reason for removing the ComboBox? You might have to reuse it here for consistency sake, as the ui currently differs from others.

Screenshot 2024-07-25 at 20 14 19
Screenshot 2024-07-25 at 20 14 32

Yomzy2223 (changes requested):

src/components/CreateWiki/EditorModals/EditWikiDetailsModal/EventsInput/index.tsx
Aliiiu: Then i will have to test others on that modal to see if they are also in a form with uncontrolled inputs. The comboBox failed the validation test because it checks for the missing type input.

Yomzy2223: The Event Dates section is the only form on the dialog. I think others work fine with the combobox. Can we modify the combobox to work with the form?

Yomzy2223 (approved):

PR #2001, opened 2024-08-06
Event Tag Link Button

Aliiiu:
# Event Tag Link Button

This PR resolves the issue (404 Error) with clicking the link button on the event tag page that previously routes to the rank page. The implemented solution routes the button to the event page with a different button content for event tag

## How should this be tested?

1. Search events
2. Open any event wiki
3. Click the event tag on the widget panel
4. On the event tag page, click the button to navigate to the event page

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/2979

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -25,7 +25,8 @@
       "recommended": true,
       "correctness": {
         "noUnusedVariables": "error",
-        "noSwitchDeclarations": "off"
+        "noSwitchDeclarations": "off",
+        "useExhaustiveDependencies": "off"
       },
       "performance": {
         "noDelete": "off"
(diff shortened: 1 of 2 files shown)


Simplypheyie96 (changes requested):
- When a tag is selected in the widget, it should filter the events to show only those that contain the selected tag. Also, there should be a clear indication on the page that these are filtered results for the specific tag, similar to how "search results for events" are displayed. 

- It's important that the tag redirects and tag sorting on the homepage behave differently to maintain a clear user experience.
Something like this:
image
for tags, it would look like this
image

Events are different from wikis, so we shouldn't confuse users.

Simplypheyie96:
> * When a tag is selected in the widget, it should filter the events to show only those that contain the selected tag. Also, there should be a clear indication on the page that these are filtered results for the specific tag, similar to how "search results for events" are displayed.
> * It's important that the tag redirects and tag sorting on the homepage behave differently to maintain a clear user experience.
>   Something like this:
> 
> image for tags, it would look like this image
> Events are different from wikis, so we shouldn't confuse users.

I realized this is a different issue and not related to this pr. Please create this comment as a new issue.

Simplypheyie96 (approved):

PR #2022, opened 2024-08-12
Reworking filter tags

Aliiiu:
# Rework Filter Tags

This PR works on fixing issues with tag filtering

## How should this be tested?

1. Go to the event page
2. Click on any of the events filter
3. Check the change in layout 

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/3007


Simplypheyie96 (changes requested):
Great work!
We have just the tags under widgets in the events details page left. They need to be reworked in a similar way too.
image

Simplypheyie96 (approved):
i think it works well now

src/components/Event/EventFilter/index.tsx
```diff
@@ -58,7 +56,7 @@ const handleFilter = (filter: Filters, dateRange?: DateRange | undefined) => {
       )
       endDate = dateFormater(lastDayNextMonth)
       break
-    case 'Custom Range':
+    case 'custom range':
```
Adebesin-Cell, line 59: why not use an enum instead of strings? 

Enums provide type safety, prevent errors from typos or invalid values, improve code readability, and offer better performance.

src/components/Event/EventInterest.tsx
```diff
@@ -70,6 +73,10 @@ const EventInterest = ({
     }
   }, [query.tags])
 
+  useEffect(() => {
```
Adebesin-Cell, line 76: What does this do?

src/pages/_app.tsx
```diff
@@ -61,9 +62,11 @@ const App: React.FC<EpAppProps> = ({
           <WagmiWrapper>
             <CSPostHogProvider>
               <PWAInstallPrompt />
-              <Layout noFooter={Component.noFooter}>
-                <Component {...pageProps} />
-              </Layout>
+              <EventContextProvider>
```
Adebesin-Cell, line 65: Hm, why are we using global context providers? Can't make it work parent -> child state?

Aliiiu: It's for a single page

Adebesin-Cell: Yep, which is why can be handled in one state, not global, I doubt we need active filter in any other page aside events page. You can prolly refactor into a hook of it'll be used in multiple places of event but having it as a global state is not right.

Softdev1 (approved):
Good work now

src/components/Event/EventInterest.tsx
Aliiiu: it clears all selected tags when the query tag is empty. adding it in the first useEffects introduce a 3-state step for the event interest toggle

src/components/Event/EventFilter/index.tsx
Aliiiu: Alright

src/pages/_app.tsx
Aliiiu: it won't work as a hook because I want to keep the same state on navigating between /event and /event/[slug].

Aliiiu: We can jump on a call so I can explain the issue better

Aliiiu: resolved it you can review it again

src/context/event.context.tsx
```diff
@@ -0,0 +1,40 @@
+import {
```
Adebesin-Cell, line 1: You can remove this file

src/components/Event/EventSearchBar.tsx
```diff
@@ -69,7 +75,7 @@ const EventSearchBar = ({
 
       const activeModifiers: ActiveModifiers = {} // assuming this is the correct type
       const mouseEvent = {} as React.MouseEvent<Element, MouseEvent> //
-      if (query.title) setSearchQuery(query.title)
+      if (query.title) setSearchQuery(String(query.title))
```
Adebesin-Cell, line 78: You can type cast query, sth like 
```ts
const { title, startDate, endDate } = router.query as { title: string; startDate: string; endDate: string }
```

src/pages/events/[slug].tsx
```diff
@@ -26,11 +30,11 @@ const EventDetailsPage = ({
   slug: string
   popularEvents: TEvents[]
 }) => {
-  const { data: countryName } = useGetIpDetailsQuery()
+  // const { data: countryName } = useGetIpDetailsQuery()
 
   const referencesRaw =
     getWikiMetadataById(event, CommonMetaIds.REFERENCES)?.value ?? '[]'
-  let references
+  let references: CiteReference[]
```
Comment (Adebesin-Cell) [inline src/pages/events/[slug].tsx:37]:
```ts

const EMPTY_ARRAY_JSON = '[]';

const getReferencesList = (event: any): CiteReference[] => {
  const referencesJsonString = getWikiMetadataById(event, CommonMetaIds.REFERENCES)?.value ?? EMPTY_ARRAY_JSON;

  try {
    const parsedReferences = JSON.parse(referencesJsonString) as CiteReference[];
    return Array.isArray(parsedReferences) ? parsedReferences : [];
  } catch (error) {
    console.error('Failed to parse references JSON:', error);
    return [];
  }
};

// In your component or function:
const referencesList = getReferencesList(event);

```

let's do this

src/pages/events/index.tsx
```diff
@@ -25,12 +26,14 @@ const EventPage = ({
...
-  const [searchActive, setSearchActive] = useState(false)
   const [isLoading, setIsLoading] = useState(false)
   const [searchDate, setSearchDate] = useState<DateRange>()
   const [searchQuery, setSearchQuery] = useState<string>('')
+  const [searchActive, setSearchActive] = useState(false)
   const router = useRouter()
-  const { data: countryName } = useGetIpDetailsQuery()
+  const currentTag = (router.query.tags as string) ? router.query.tags : ''
+  const activeTagFilter = currentTag.length > 0
+  // const { data: countryName } = useGetIpDetailsQuery()
```
Adebesin-Cell, line 36: remove unused code

src/pages/events/index.tsx
```diff
@@ -25,12 +26,14 @@ const EventPage = ({
...
 }) => {
   const [eventData, setEventData] = useState<TEvents[]>(events)
-  const [searchActive, setSearchActive] = useState(false)
   const [isLoading, setIsLoading] = useState(false)
   const [searchDate, setSearchDate] = useState<DateRange>()
   const [searchQuery, setSearchQuery] = useState<string>('')
+  const [searchActive, setSearchActive] = useState(false)
   const router = useRouter()
-  const { data: countryName } = useGetIpDetailsQuery()
+  const currentTag = (router.query.tags as string) ? router.query.tags : ''
```
Adebesin-Cell, line 34: type mismatch? tags suggest ``string[]``

src/pages/events/index.tsx
```diff
@@ -25,12 +26,14 @@ const EventPage = ({
...
+  const activeTagFilter = currentTag.length > 0
```
Adebesin-Cell, line 35: Improve variable name so it gives a nicer description of what the boolean does, perhaps ``isThereCurrentTag`` based on context should be ``hasActiveTag``

src/utils/event.utils.ts
```diff
@@ -11,6 +12,8 @@ export const isEventLocation = (location: Location | Location[]) => {
   return true
 }
 
+export type SetState<T> = Dispatch<SetStateAction<T>>
```
Adebesin-Cell, line 15: Hm, why defining this?

src/components/Event/EventInterest.tsx
```diff
@@ -11,16 +12,16 @@ const EventInterest = ({
...
 }) => {
   const router = useRouter()
   const { t } = useTranslation('event')
   const { query, pathname } = router
-  const [selectedTags, setSelectedTags] = useState<string[]>([])
+  const [selectedInterest, setSelectedInterest] = useState<string[]>([])
 
   const toggleTag = (tag: string) => {
-    setSelectedTags((prevSelected) => {
+    setSelectedInterest((prevSelected) => {
```
Adebesin-Cell, line 24: What does Interest mean?

src/components/Event/EventInterest.tsx
```diff
@@ -62,11 +63,15 @@ const EventInterest = ({
...
   useEffect(() => {
-    const tagsFromQuery = query.tags
-    if (tagsFromQuery) {
-      const tagsArray =
-        typeof tagsFromQuery === 'string' ? [tagsFromQuery] : tagsFromQuery
-      setSelectedTags(tagsArray)
+    const interestFromQuery = query.interest
+    if (interestFromQuery) {
+      const interestArray =
+        typeof interestFromQuery === 'string'
```
Adebesin-Cell, line 69: Idg, do we have an array of interests in url then we parse or we are doing some parses before we achieve a field of interests?

src/utils/event.utils.ts
Aliiiu: it is a type for setState actions passed as props to other components. Instead of repeating the same type every where I created a custom type that works for every case

src/pages/events/index.tsx
Aliiiu: It is to be revised and tested well later. We are still using it

src/pages/events/index.tsx
Aliiiu: The expected query in this case is a string. The logic here is to filter events by tags when you click any tag in the event summary widget on the event details page. Clicking any of the tags take you to the event page.

src/components/Event/EventInterest.tsx
Aliiiu: Event interest in the event interest cards

src/components/Event/EventInterest.tsx
Aliiiu: Multiple interests can be selected, then they are parsed to an array. The query is expecting [string!]

src/utils/event.utils.ts
Adebesin-Cell: I think it makes sense to allow for the repetition so it doesn't get complex to read code 🙏

src/components/Event/EventInterest.tsx
Adebesin-Cell: Do you have a picture perhaps?

src/components/Event/EventInterest.tsx
Adebesin-Cell: you mean filters? I'm confused how does it get passed to query?

src/pages/events/index.tsx
Adebesin-Cell: Reads better as tag, tags gives sth like plural

src/pages/events/index.tsx
Adebesin-Cell: If current tag, then you can just check.
```ts
	const { tag } = router.query as {tag?: string}
 const activeTag = tab || null
```

then remove the activeTagFilter condition, and use activeTag

src/components/Event/EventInterest.tsx
Aliiiu: image

Adebesin-Cell: Thanks!

src/components/Event/EventInterest.tsx
Adebesin-Cell: we can name them interests and setSelectedInterests, also if there's a way to push to the .interests in query would be nicer like we have https://iq-wiki-git-reworking-filter-tags-prediqt.vercel.app/events/?interests=BinanceSmartChain&Blockchains&Protocols, then we know we have to parse all into an array

Aliiiu: Using the picture attached to the previous conversation as a reference. To get events that have some selected tags [in the UI it is interests] we run a query that accepts tagIds:[string!]. When you click any of the interests it adds a state to the URL which you can get from useRouter() as a sting if it is a single param you add or an array if you add multiple params.

image

Adebesin-Cell: ``interestArray`` can be 

we can do the interest parsing here to array from the stringified query
```ts
const interestQueryParam = query.interest;

const parsedInterests = parseInterestQueryParam(interestQueryParam);

function parseInterestQueryParam(param: string | string[] | undefined): string[] {
  if (!param) {
    return [];
  }
  return Array.isArray(param) ? param : [param];
}
```

src/pages/events/index.tsx
```diff
@@ -91,7 +94,7 @@ const EventPage = ({
               />
             </div>
             <div className="grid md:grid-cols-2 xl:grid-cols-1 gap-10 md:gap-4 lg:gap-10">
-              <NearbyEventFilter countryName={countryName ?? ''} />
+              {/* <NearbyEventFilter countryName={countryName ?? ''} /> */}
```
Adebesin-Cell, line 97: remove unused line

src/pages/events/index.tsx
```diff
@@ -3,19 +3,20 @@ import TrendingEvent from '@/components/Event/TrendingEvent'
 import EventBanner from '@/components/Event/EventBanner'
 import EventInterest from '@/components/Event/EventInterest'
 import EventSearchBar from '@/components/Event/EventSearchBar'
-import NearbyEventFilter from '@/components/Event/NearbyEventFilter'
+// import NearbyEventFilter from '@/components/Event/NearbyEventFilter'
```
Adebesin-Cell, line 6: remove unused line

src/pages/events/[slug].tsx
```diff
@@ -26,11 +30,11 @@ const EventDetailsPage = ({
   slug: string
   popularEvents: TEvents[]
 }) => {
-  const { data: countryName } = useGetIpDetailsQuery()
+  // const { data: countryName } = useGetIpDetailsQuery()
```
Comment (Adebesin-Cell) [inline src/pages/events/[slug].tsx:33]: remove unsed line

src/components/Event/EventList.tsx
```diff
@@ -65,9 +74,11 @@ const EventList = ({
 
   return (
     <div className="flex flex-col flex-1 gap-5">
-      {searchActive && (
+      {(searchActive || activeTagFilter) && (
         <span className="flex flex-col items-start">
-          <h1 className="font-semibold">{t('searchResult')}</h1>
+          <h1 className="font-semibold">
+            {activeTagFilter ? 'Event Filter Results' : t('searchResult')}
```
Adebesin-Cell, line 80: Ensure to add translations to Event Filter Results text

src/components/Event/EventSearchBar.tsx
Aliiiu: Nice, some other variable will share the same names with the ones you declared here and they are in the same scope. it will raise a conflict so I used this

`const query = router.query as {
      title: string
      startDate: string
      endDate: string
    }`

src/pages/events/[slug].tsx
Comment (Adebesin-Cell) [reply src/pages/events/[slug].tsx:37]: It reads better to have the function and UI separate instead of having the parsing of references, we could have a function that does that, I think it can also be used in wiki again.

src/pages/events/index.tsx
Adebesin-Cell: In this code or another pr?

Aliiiu: Part of the issue is to hide nearByEvent component

src/components/Event/EventInterest.tsx
```diff
@@ -62,12 +63,17 @@ const EventInterest = ({
   }
 
   useEffect(() => {
-    const tagsFromQuery = query.tags
-    if (tagsFromQuery) {
-      const tagsArray =
-        typeof tagsFromQuery === 'string' ? [tagsFromQuery] : tagsFromQuery
-      setSelectedTags(tagsArray)
+    const interestQueryPaaram = query.interest
```
Adebesin-Cell, line 66: ``interestQueryParam``

src/components/Event/EventInterest.tsx
```diff
@@ -62,12 +63,17 @@ const EventInterest = ({
...
+
+    function parseInterestQueryParam(
+      param: string | string[] | undefined,
+    ): string[] {
+      if (!param) {
+        return []
+      }
+      return Array.isArray(param) ? param : [param]
     }
+    setSelectedInterest(parseInterestQueryParam(interestQueryPaaram))
```
Adebesin-Cell, line 76: better to always call and assign to a value before use state

I mean assign parseInterestQueryParam(interestQueryPaaram) to a variable then use the variable in state setter

src/components/Event/EventInterest.tsx
```diff
@@ -62,12 +63,17 @@ const EventInterest = ({
...
 
   useEffect(() => {
-    const tagsFromQuery = query.tags
-    if (tagsFromQuery) {
-      const tagsArray =
-        typeof tagsFromQuery === 'string' ? [tagsFromQuery] : tagsFromQuery
-      setSelectedTags(tagsArray)
+    const interestQueryPaaram = query.interest
+
+    function parseInterestQueryParam(
```
Adebesin-Cell, line 68: Move this to event.utils

src/components/Event/EventSearchBar.tsx
Adebesin-Cell: you can rename them to fit if such might happen, maybe ``{title: filterTitle}``

src/pages/events/[slug].tsx
```diff
@@ -1,19 +1,23 @@
...
-import NearbyEventFilter from '@/components/Event/NearbyEventFilter'
+// import NearbyEventFilter from '@/components/Event/NearbyEventFilter'
 import PopularEventFilter from '@/components/Event/PopularEventFilter'
 import { EventHeader } from '@/components/SEO/Event'
 import RelatedMediaGrid from '@/components/Wiki/WikiPage/InsightComponents/RelatedMedia'
 import WikiReferences from '@/components/Wiki/WikiPage/WikiReferences'
-import { TEvents, getPopularEvents } from '@/services/event'
-import { useGetIpDetailsQuery } from '@/services/location'
+import { type TEvents, getPopularEvents } from '@/services/event'
+// import { useGetIpDetailsQuery } from '@/services/location'
```
Comment (Adebesin-Cell) [inline src/pages/events/[slug].tsx:10]: Unused imports

src/pages/events/[slug].tsx
```diff
@@ -26,18 +30,22 @@ const EventDetailsPage = ({
   slug: string
   popularEvents: TEvents[]
 }) => {
-  const { data: countryName } = useGetIpDetailsQuery()
+  // const { data: countryName } = useGetIpDetailsQuery()
 
-  const referencesRaw =
-    getWikiMetadataById(event, CommonMetaIds.REFERENCES)?.value ?? '[]'
-  let references
+  const getReferencesList = (event: any): CiteReference[] => {
```
Comment (Adebesin-Cell) [inline src/pages/events/[slug].tsx:35]: move function to utils, it leaves a good code organization

src/pages/events/index.tsx
```diff
@@ -25,12 +25,14 @@ const EventPage = ({
...
   const [eventData, setEventData] = useState<TEvents[]>(events)
-  const [searchActive, setSearchActive] = useState(false)
   const [isLoading, setIsLoading] = useState(false)
   const [searchDate, setSearchDate] = useState<DateRange>()
   const [searchQuery, setSearchQuery] = useState<string>('')
+  const [searchActive, setSearchActive] = useState(false)
   const router = useRouter()
-  const { data: countryName } = useGetIpDetailsQuery()
+  const currentTag = (router.query.tag as string) ? router.query.tag : ''
+  const hasActiveTags = currentTag.length > 0
```
Adebesin-Cell, line 34: Hm, since it's a string instead of checking the length we can add a null as value instead of empty string, then we do null check for hasActive	Tag, except if tag is an array of selection, then we can rename it to currentTags

Adebesin-Cell:
Fix build errors, but I think aside that code looks good

Adebesin-Cell (approved):

PR #2039, opened 2024-08-15
Implemented distinct styling for active links

Solexi:
# Add distinct style for active nav items

_Implements a distinct style for active navigation items in the Navbar component._

## How should this be tested?

1. Run the application.
2. Navigate through different pages.
3. Check that the active navigation items have a distinct style.

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/3037


Tolu-Mals:
Nice work. I see you implemented an active state for the sub-menus as well. It's not very apparent, use the same styles for the hover state as the active state. Everything else looks good.

SidharthK2:
> Nice work. I see you implemented an active state for the sub-menus as well. It's not very apparent, use the same styles for the hover state as the active state. Everything else looks good.

cc @Simplypheyie96

Simplypheyie96:
> > Nice work. I see you implemented an active state for the sub-menus as well. It's not very apparent, use the same styles for the hover state as the active state. Everything else looks good.
> 
> cc @Simplypheyie96

I think we can have the same active states for md/sm breakpoints as well in the menu.

Tolu-Mals (approved):

Simplypheyie96 (approved):

PR #2040, opened 2024-08-16
Preload state for images

invisiblemask:
# Add preload states for images

_Images don't have a preload state when selecting category tabs, they appear empty when other content loads, preload skeleton state has been added to ensure a good user experience when viewing each category tab_

## How should this be tested?

1. On the homepage, scroll to the categories section
2. Click on each tab to see the preload state

closes https://github.com/EveripediaNetwork/issues/issues/3057


Tolu-Mals:
I personally don't like the gradient animation skeleton. I think something like this looks better 

https://tailwindcss.com/docs/animation#pulse

 @Simplypheyie96  what do you think?
 
https://github.com/user-attachments/assets/1cb36b46-dde8-40a6-8f00-d2bf5f0b7226

Simplypheyie96:
> I personally don't like the gradient animation skeleton. I think something like this looks better
> 
> https://tailwindcss.com/docs/animation#pulse
> 
> @Simplypheyie96 what do you think?
> 
>  Screen.Recording.2024-08-16.at.10.00.43.AM.mov

This looks like a gradient too but a lot subtler. it would look good. 
 @invisiblemask The issue is still there. the blank card before the image finally shows.

invisiblemask:
> > I personally don't like the gradient animation skeleton. I think something like this looks better
> > https://tailwindcss.com/docs/animation#pulse
> > @Simplypheyie96 what do you think?
> > Screen.Recording.2024-08-16.at.10.00.43.AM.mov
> 
> This looks like a gradient too but a lot subtler. it would look good. @invisiblemask The issue is still there. the blank card before the image finally shows.

Yeah, I amm currently fixing it, so are we going with @Tolu-Mals  suggestion for the gradient?

Simplypheyie96:
> Yeah, I amm currently fixing it, so are we going with @Tolu-Mals suggestion for the gradient?

yeah. let's do that. 
@Softdev1 you should create an issue to update other wiki cards with this preload state too.

Softdev1:
> > Yeah, I amm currently fixing it, so are we going with @Tolu-Mals suggestion for the gradient?
> 
> yeah. let's do that. @Softdev1 you should create an issue to update other wiki cards with this preload state too. I don't know if we can do that on this pr
Let's use  pull request to resolve the preload state issues across board. No need to create another issue, I would rather list them out as tasks under the same issue.

Simplypheyie96 (approved):

PR #2047, opened 2024-08-16
Rearrange categries 4 col

invisiblemask:
# Rearrangement of Categories card from 3 columns to 4 

_This PR fixes the issue of replacing 3 columns with 4 for the categories section, this makes images that are pixelated clearer_

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/3065


src/components/Categories/CategoriesCard.tsx
```diff
@@ -45,7 +45,7 @@ export default function CategoriesCard({ wiki }: CategoriesCardProps) {
           {shortenText(wiki.title, 30)}
         </h1>
         <p className="dark:text-alpha-800 text-gray-600 text-xs lg:text-sm">
-          {shortenText(wiki.summary, 80)}
+          {shortenText(wiki.summary, 50)}
```
kesar, line 48: not big deal but code would look much better if we extract these 50, 80, 30 magic numbers into constants on top of the component and use it there 👍🏻

src/components/Landing/CategoriesList.tsx
```diff
@@ -78,14 +78,14 @@ const CategoriesList = ({ categories }: CategoriesListProps) => {
...
                     (category) =>
-                      category?.id === allCategory.id && (
+                      category.id === allCategory.id && (
                         <div
                           key={category.id}
-                          className="grid grid-cols-1 md:grid-cols-2 xl:grid-cols-3 gap-6"
+                          className="grid grid-cols-1 md:grid-cols-2 xl:grid-cols-4 gap-6"
                         >
-                          {category.wikis.slice(0, 6).map((wiki) => (
+                          {category.wikis.slice(0, 8).map((wiki) => (
```
kesar, line 88: same wth this number, probably extract and const in the beginning to easily spot and replace

kesar:
great work, just some small suggestions

src/components/Categories/CategoriesCard.tsx
```diff
@@ -31,7 +31,7 @@ export default function CategoriesCard({ wiki }: CategoriesCardProps) {
       href={`/wiki/${wiki.id}`}
       className="flex flex-col gap-2 border dark:border-gray-700 border-gray-200 rounded-xl pb-2 hover:scale-105 transition-transform duration-300 ease-in-out"
     >
-      <div className="h-80">
+      <div className="aspect-auto">
```
kesar, line 34: not sure how this work, but lets ensure this doesnt make things weird in some of the different screen sizes

src/components/Categories/CategoriesCard.tsx
invisiblemask: Okay sir

src/components/Categories/CategoriesCard.tsx
invisiblemask: I am already testing that

Simplypheyie96 (changes requested):
1. it's for home and categories page, so we can't close this out yet.
2. There is inconsistency with the placement of the article author and i am just noticing that the "date edited is missing".
image

3. just like the design, we would have to adjust the font size of the wiki title and subtitles on the card since the cards are smaller now.

invisiblemask:
> 1. it's for home and categories page, so we can't close this out yet.
> 
> 2. There is inconsistency with the placement of the article author and i am just noticing that the "date edited is missing".
> 
> image
> 
> 
> 
> 3. just like the design, we would have to adjust the font size of the wiki title and subtitles on the card since the cards are smaller now.

Do you mean changing the wiki cards in categories page to 4 columns too?

The author name is not consistent because the wiki summary length is very small and that's because it's a test wiki, the summary length for the actual wikis are longer, but I will adjust that in case there's any scenario where we would have short wiki summary length and also add the date edited too

I will check the design and update the font sizes

Simplypheyie96:
> Do you mean changing the wiki cards in categories page to 4 columns too?

It was added as part of the issue. @Softdev1 please confirm again if we are changing the categories page into 4 columns.
image

Softdev1:
> > Do you mean changing the wiki cards in categories page to 4 columns too?
> 
> It was added as part of the issue. @Softdev1 please confirm again if we are changing the categories page into 4 columns. image

Yes we are doing it on both the homepage and the category page too

Simplypheyie96:
> Yes we are doing it on both the homepage and the category page too

To avoid pr conflict, @Yomzy2223  can work on the the categories page since he is already working on the categories page.

Simplypheyie96 (approved):

Tolu-Mals:
When the user doesn't have a username, it should show the address instead of being blank

Screenshot 2024-08-20 at 9 39 40 AM

Reference - Activity Card
Screenshot 2024-08-20 at 9 40 16 AM

Softdev1 (approved):

Tolu-Mals (approved):

PR #2059, opened 2024-08-21
Rankpage Search Functionality

Damola18:
# Rankpage Search Functionality

This PR:
- adds search functionality to all the tabs on the rankapage which allow users to easily search for tokens, stablecoins, nfts and other rank items.
- improves the rank pagination UX by hiding the prev/next buttons on the first/last page respectively in order to avoid showing negative table entries on few occasions

## How should this be tested?

1. Go to the rank page
2. Search for any item within the table across all 5 tabs
3. Use the pagination buttons to ensure negative table entries are not shown as well.

## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/3039


Damola18:
Just a side note for observation, @Simplypheyie96 @SidharthK2 ,  the 5 sections have now been refactored to use a single reusable component instead of 5 different tabs, so the description text for the founders tab uses a single color.

Simplypheyie96 (changes requested):
- Let's limit the heading texts to a maximum of two lines to maintain consistency in the placement of the search bar and prevent layout shifts
image

- Also, the search bar should function across all content, not that each tab has it's own search. (is that not feasible?)
-
- We shouldn't force all content into a small space, as this can lead to display issues, as seen below. Let's give appropriate paddings.
image

Damola18:
> * Let's limit the heading texts to a maximum of two lines to maintain consistency in the placement of the search bar and prevent layout shifts
> 
> image
> * Also, the search bar should function across all content, not that each tab has it's own search. (is that not feasible?)
> * We shouldn't force all content into a small space, as this can lead to display issues, as seen below. Let's give appropriate paddings.
> 
> image

Each tabs has its own designated data table with each values fetched from their respective endpoints, so I don't think that's feasible actually.

Damola18:
> > Each tabs has its own designated data table with each values fetched from their respective endpoints, so I don't think that's feasible actually.
> 
> Alright. But we can make changes to rest of the issues.

sure, will do that right away.

Simplypheyie96:
Looks a lot better. 
We need to update the layout issue on mobile and remove gradients too. And that would be all.

src/components/Rank/RankTabPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+    setCurrentPage(1) // Reset to first page when search query changes
+  }
+
+  const memoizedItems = useMemo(() => {
+    if (searchQuery.trim().length < SEARCH_LENGTH) return items
+
+    return items.filter((item) =>
+      item.title.toLowerCase().includes(searchQuery.toLowerCase()),
+    )
+  }, [items, searchQuery])
```
SidharthK2, line 98: Consider more descriptive naming

src/components/Rank/RankTabPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  // const showResults = memoizedItems.length > 0 || isFetching;
```
SidharthK2, line 100:
```suggestion
  
```

SidharthK2 (commented):
App is crashing when I navigate to rank page

Damola18:
> App is crashing when I navigate to rank page

Screenshot 2024-08-22 at 10 27 32

Yeah, this is what breaks the page, tracing where that came from.

Simplypheyie96 (approved):

PR #2084, opened 2024-08-30
fixes create wiki bug

Adebesin-Cell:
# fixes create wiki bug

This PR adds the auth check for create wiki in get server side props and also refactored the code

Fixes https://github.com/EveripediaNetwork/issues/issues/3102


Tolu-Mals (approved):
Great pr ✨

PR #2086, opened 2024-09-02
Integrates IQ Login Package

Royal-lobster:
# changes
- removes the existing login, magic and wamgi setup
- updates the wagmi version and adds patches to the outdated hook uses
- adds login page from iq-login package


Royal-lobster:
@SidharthK2 @Yadheedhya06 Help checking out the login in this pr 

Test out these:
1. Login with metamask, google, email
1. Create wiki
1. Edit wiki

Painpoints before to keep in mind while testing:
1. Not logging in / irresponsive login buttons sometimes
1. Switch network won't work for some times
1. Random loggouts on wiki publish button click

If it works fine, ill proceed to merge it so we can check with editors if they are facing any issues with the login.

PR #2101, opened 2024-09-03
Rank pagination issue

Aliiiu:
# Rank Pagination issues

We are currently using the wrong offset to fetch rank data on home page and rank page. This result in having some missing data in the home page and wrong pagination in the rank. The pagination adds one element to the current list which is a wrong behavior. This PR also resolves issues with missing images for fetched rank data.

## How should this be tested?

1. Go to rank page
2. Use the next button to move from one list to the other

## Notes or observations

<summary>
 <details>Cryptocurrency rank page (Prod db)</details>
image

</summary>
## Linked issues

closes https://github.com/EveripediaNetwork/issues/issues/ISSUE_NUMBER_GOES_HERE


AshakaE:
Pagination looks good, and top tokens per list also, but the page number stops at 3 
image

invisiblemask (commented):
When I click on the pagination button for cryptocurrencies table it shows the next data then I click on another tab and go back to the cryptocurrencies tab the data remains the same but the pagination button returns to 1

https://github.com/user-attachments/assets/bfd15bfa-b830-4a20-842a-cbb957ba9984

